PR #127, opened 2024-04-17
Fix issue with generic errors being returned with gating errors

kemmerle:
## Description and Context
During our latest bug bashes, folks were receiving generic errors while attempting the `hs project upload` command, particularly when their accounts had gating issues. 

With @joe-yeager's help (he gave me access to his account, so I could reproduce the error), I found that we were destructuring the error object incorrectly and therefore not passing the correct status code to the switch statement in the `getAxiosErrorWithContext` function. 

## Screenshots

BEFORE (Yucky, generic error):

Screenshot 2024-04-17 at 4 56 53 PM

AFTER (Beautiful, descriptive error):

Screenshot 2024-04-17 at 4 57 25 PM

## TODO

- [ ] Address any feedback

## Who to Notify

@joe-yeager @brandenrodgers

### DIFF
--- a/errors/apiErrors.ts
+++ b/errors/apiErrors.ts
@@ -1,4 +1,4 @@
-import { AxiosError } from 'axios';
+import { AxiosError, AxiosResponse } from 'axios';
 import {
   GenericError,
   AxiosErrorContext,
@@ -130,7 +130,7 @@ export function getAxiosErrorWithContext(
   error: AxiosError<any>,
   context: AxiosErrorContext = {}
 ): Error {
-  const { status } = error;
+  const { status } = error.response as AxiosResponse;
   const method = error.config?.method as HttpMethod;
   const { projectName } = context;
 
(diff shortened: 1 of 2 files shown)


joe-yeager (approved):

errors/apiErrors.ts
```diff
@@ -130,7 +130,10 @@ export function getAxiosErrorWithContext(
   error: AxiosError<any>,
   context: AxiosErrorContext = {}
 ): Error {
-  const { status } = error;
+  let status;
```
brandenrodgers, line 133: You could shorten this to `const status = error.response?.status`. Only mentioning it because it's slightly more consistent with what we're already doing in other spots in this file ([example](https://github.com/HubSpot/hubspot-local-dev-lib/pull/127/files#diff-9d5b4d724ee9d55d1fa4e316c00ad65c422dc4365c603388afd8f012d3327158R34))

brandenrodgers (approved):
Just one small suggestion to shorten things and make it more consistent, but otherwise lgtm

PR #130, opened 2024-05-01
Feature: apps migration and cloning (Sister PR in hubspot-cli) 

kemmerle:
## Description and Context
In this PR, I'm adding support for public apps migration and cloning. Because the necessary endpoints are still being worked on, I'm currently scaffolding out the commands using test data. I will update this PR, as endpoints become available.

There is a [sister PR](https://github.com/HubSpot/hubspot-cli/pull/1052) in hubspot-cli.

## Screenshots

`hs project create --migrateApp`:

Happy path: 
Screenshot 2024-05-09 at 4 03 36 PM

Sad path (No errors being returned yet):
Screenshot 2024-05-09 at 4 05 45 PM

## TESTING

The `hs project migrate-app` command is now ready for testing. You will need to go through the `yarn link` process to take advantage of the API endpoints in the local-dev-lib library (See PR linked above). Make sure that you switch to the correct branch before testing. 

Currently, the BE does not create and link an app, but we do create a new project in your developer account. I've also added downloading the project files to your local machine, so make sure to check that, as well.  

IMPORTANT CAVEAT: Currently, no errors are being returned, so we're just returning a generic error message. 

## TODO

 

- [x] Update endpoints as they become available
- [x] Address feedback
- [x] Test out different scenarios

## Who to Notify

@brandenrodgers @camden11 @joe-yeager @mendel-at-hubspot


api/projects.ts
```diff
@@ -288,3 +289,24 @@ export async function fetchDeployWarnLogs(
...
+    url: `${MIGRATIONS_API_PATH}/migrations`,
+    data: {
+      componentId: appId,
+      componentType: 'PUBLIC_APP_ID',
+      projectName,
+    },
+  });
+}
+
+export async function checkMigrationStatus(accountId: number, id: number) {
```
brandenrodgers, line 308: Do these api requests have return types that we can include?

lib/archive.ts
```diff
@@ -73,7 +73,7 @@ async function copySourceToDest(
   { sourceDir, includesRootDir = true }: CopySourceToDestOptions = {}
 ): Promise<boolean> {
   try {
-    logger.log(i18n(`${i18nKey}.copySourceToDest.init`));
+    logger.debug(i18n(`${i18nKey}.copySourceToDest.init`));
```
brandenrodgers, line 76: Why were these changed to debug logs?

types/App.ts
```diff
@@ -0,0 +1,26 @@
...
+  supportUrl?: string;
+  supportEmail?: string;
+  supportPhone?: string;
+  extensionIconUrl?: string;
+  isAdvancedScopesSettingEnabled: boolean;
+};
+
+export type FetchAppsResponse = {
+  results: Array<App>;
+};
```
camden11, line 26: I made a `types/app` file recently that this could be moved to. Also made a type for `PublicApp` - you may be able to use that or make some kind of generic `App` type and then extend it

camden11: https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/types/Apps.ts

lib/archive.ts
kemmerle: The logs were appearing unexpectedly when I was downloading the files for the new post-migration project. I was thinking of adding a boolean to switch between debug/log but wasn't sure--meant to leave a note with a question but then John was angsty to get the changes out and I forgot 😬

types/App.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/130/commits/fbd7671efc9a0dafef58bd063c610257dc976f99.

We were importing the same endpoint, so I deleted my work and opted for yours 👍

api/projects.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/130/commits/d19b22a17f35d945c8645df41b3e618bc3f8b3f8

camden11 (approved):

lib/archive.ts
kemmerle: Discussed this with @camden11 in a Huddle, and he suggested we add a `hideLogs` boolean, so we don't cause any issues with other commands that use these functions.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/130/commits/d86aa5e928b94628db2a49eca86371ba33e2db70

PR #131, opened 2024-05-06
Update sandbox types, replace polling API endpoint

adamawang:
## Description and Context

This PR replaces an endpoint that sandboxes was using to poll for sync updates with a much simpler and more performant one. No changes are needed in hubspot-cli as the data that [pollSyncTaskStatus](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/sandboxes.js#L265-L326) expects does not change with this endpoint swap. 

I've also gone ahead and cleaned up some types that were ported over in a slightly weird state. 

To run this locally, navigate to your local instance of `hubspot-local-dev-lib` and run `yarn run local-dev`. Then go into your local instance of `hubspot-cli` and run `yarn link @hubspot/local-dev-lib` to link the local version. You can now run `yarn hs sandbox sync` to verify that everything still works. There should be no changes to what is displayed. 

## Screenshots

No changes to the polling UI, tested and is still same as before
Screenshot 2024-05-06 at 17 17 25

## TODO

n/a

## Who to Notify

@shannon-lichtenwalter

### DIFF
--- a/api/sandboxSync.ts
+++ b/api/sandboxSync.ts
@@ -1,9 +1,9 @@
 import http from '../http';
 import {
   InitiateSyncResponse,
-  Task,
-  SyncTask,
   FetchTypesResponse,
+  TaskRequestData,
+  SyncTaskStatusType,
 } from '../types/Sandbox';
 import { SANDBOX_TIMEOUT } from '../constants/api';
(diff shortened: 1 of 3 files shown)


camden11 (approved):
Looks good + tested and works!

PR #134, opened 2024-05-09
Changes for enabling VSCE migration, update module creation

jsines:
## Description and Context
Adds a few things VSCode extension relies on from `cli-lib` to enable migration

Also updates module creation in `local-dev-lib` to use `content_types` instead of host template


lib/cms/templates.ts
```diff
@@ -79,3 +79,46 @@ export async function createTemplate(
     filePath
   );
 }
+
+export const TEMPLATE_TYPES = {
```
j-malt, line 83: Where's this list from?

j-malt (approved):

lib/cms/templates.ts
jsines: Yep directly from https://github.com/HubSpot/cli-lib/blob/f7d547abbde458d4c394deab48b1ac0bb5376bef/lib/constants.js#L130 !

PR #136, opened 2024-05-22
Add code of conduct

kemmerle:
## Description and Context
In our efforts to deprecate the `cli-lib` library in favor of `local-dev-lib`, we are adding a code of conduct for HubSpot developers who would like to contribute here. 

Based on our discussions in Slack, I've based this code of conduct on the developer relation team's [CoC](https://github.com/hubspotdev/.github/blob/main/CODE_OF_CONDUCT.md) which @ajlaporte was kind enough to link me to. 

My main questions now are:

1) What kind of promises would we like to make to our developer community? 
2) Should we provide a form for developers to report any violations, as the DevRel team does? 

## TODO

- [x] Address any feedback

## Who to Notify

@joe-yeager @camden11 @brandenrodgers

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,60 @@
+# HubSpot Local Dev Lib Code of Conduct
+
+## Our Pledge to you, our community of developers
+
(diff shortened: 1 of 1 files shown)


camden11 (commented):
Hmm it does seem like if we're making all these promises, we should have a way to let people report violations. I think a form would make sense. I wonder if we could just use the existing DevRel form or if we need our own

camden11 (approved):
LGTM!

PR #137, opened 2024-05-23
Add function to hit test portal installs endpoint

camden11:
## Description and Context
This adds functionality to hit a new endpoint that returns the number of installs a public app has in developer test account. We'll use this on the CLI to determine if an app has installs that _aren't_ on test accounts (see https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/571)

## Who to Notify

### DIFF
--- a/api/appsDev.ts
+++ b/api/appsDev.ts
@@ -1,5 +1,5 @@
 import http from '../http';
-import { PublicApp } from '../types/Apps';
+import { PublicApp, DeveloperTestAccountInstallData } from '../types/Apps';
 
 const APPS_DEV_API_PATH = 'apps-dev/external/public/v3';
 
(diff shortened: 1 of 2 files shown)


api/appsDev.ts
```diff
@@ -16,3 +19,12 @@ export async function fetchPublicAppsForPortal(
 
   return resp ? resp.results : [];
 }
+
+export function fetchPublicAppDeveloperTestAccountInstallData(
+  appId: number,
+  accountId: number
+): Promise<PublicAppDeveloperTestAccountInstallData> {
+  return http.get<PublicAppDeveloperTestAccountInstallData>(accountId, {
```
joe-yeager, line 27: I know this is a very simple method, but I think a unit test or two to verify the contract might be beneficial.

joe-yeager (commented):
LGTM, just one comment about possibly adding a test

joe-yeager (approved):

brandenrodgers (approved):

PR #139, opened 2024-05-29
fix: Add keepAlive to axios agent

joe-yeager:
## Description and Context

We were getting `ENOMEM` errors from `axios` because it appears to have a [memory leak starting in](https://github.com/axios/axios/issues/5641) `1.4.0` and in versions as late as `1.6.8`. The leak is caused by event listeners not being closed when http calls complete.  A solution to fix or at least limit the impact is to set `keepAlive` to true and limit the number of sockets which causes `axios` to reuse the connections instead of creating new ones.  

As an aside, we **_might_** notice some minor performance benefits as well on request heavy operations because we will be reusing connections.

[More context on connection management](https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x)

Changes:
- Pass custom `http` and `https` agents to `axios`
- Set `keepAlive` to true to allow connections to be kept open
- Set a limit on the number of total connections
- Set a limit on the number of connections per domain

### DIFF
--- a/http/getAxiosConfig.ts
+++ b/http/getAxiosConfig.ts
@@ -7,8 +7,23 @@ import { AxiosRequestConfig } from 'axios';
 import https from 'https';
 import http from 'https';
 
-const httpAgent = new http.Agent({ keepAlive: true });
-const httpsAgent = new https.Agent({ keepAlive: true });
+// Total number of sockets across all hosts
+const MAX_TOTAL_SOCKETS = 50;
+
+// Total number of sockets per each host
+const MAX_SOCKETS_PER_HOST = 10;
+
+const httpAgent = new http.Agent({
+  keepAlive: true,
+  maxTotalSockets: MAX_TOTAL_SOCKETS,
+  maxSockets: MAX_SOCKETS_PER_HOST,
+});
+
+const httpsAgent = new https.Agent({
+  keepAlive: true,
+  maxTotalSockets: MAX_TOTAL_SOCKETS,
+  maxSockets: MAX_SOCKETS_PER_HOST,
+});
 
 export const USER_AGENTS: { [key: string]: string } = {
   'HubSpot Local Dev Lib': version,
@@ -38,7 +53,6 @@ export function getAxiosConfig(
   const { env, localHostOverride, headers, ...rest } = options;
   const { httpTimeout, httpUseLocalhost } =
     getAndLoadConfigIfNeeded() as CLIConfig;
-
   return {
     baseURL: getHubSpotApiOrigin(
       env,
--- a/http/getAxiosConfig.ts
+++ b/http/getAxiosConfig.ts
@@ -5,7 +5,7 @@ import { AxiosConfigOptions } from '../types/Http';
 import { CLIConfig } from '../types/Config';
 import { AxiosRequestConfig } from 'axios';
 import https from 'https';
-import http from 'https';
+import http from 'http';
 
 // Total number of sockets across all hosts
 const MAX_TOTAL_SOCKETS = 25;
--- a/http/getAxiosConfig.ts
+++ b/http/getAxiosConfig.ts
@@ -4,6 +4,11 @@ import { getHubSpotApiOrigin } from '../lib/urls';
 import { AxiosConfigOptions } from '../types/Http';
 import { CLIConfig } from '../types/Config';
 import { AxiosRequestConfig } from 'axios';
+import https from 'https';
+import http from 'https';
+
+const httpAgent = new http.Agent({ keepAlive: true });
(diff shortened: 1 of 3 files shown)


types/Utils.ts
```diff
@@ -9,5 +9,5 @@ type Join<K, P> = K extends string | number
 export type Leaves<T> = [10] extends [never]
   ? never
   : T extends object
-  ? { [K in keyof T]-?: Join<K, Leaves<T[K]>> }[keyof T]
-  : '';
+    ? { [K in keyof T]-?: Join<K, Leaves<T[K]>> }[keyof T]
```
joe-yeager, line 12: Unrelated, but prettier was not happy with this file.

http/getAxiosConfig.ts
```diff
@@ -4,6 +4,26 @@ import { getHubSpotApiOrigin } from '../lib/urls';
...
+const MAX_TOTAL_SOCKETS = 25;
+
+// Total number of sockets per each host
+const MAX_SOCKETS_PER_HOST = 5;
+
+const httpAgent = new http.Agent({
+  keepAlive: true,
+  maxTotalSockets: MAX_TOTAL_SOCKETS,
+  maxSockets: MAX_SOCKETS_PER_HOST,
+});
```
joe-yeager, line 20: [Docs on these values](https://nodejs.org/api/http.html#new-agentoptions)

http/getAxiosConfig.ts
```diff
@@ -4,6 +4,26 @@ import { getHubSpotApiOrigin } from '../lib/urls';
...
 import { CLIConfig } from '../types/Config';
 import { AxiosRequestConfig } from 'axios';
+import https from 'https';
+import http from 'https';
+
+// Total number of sockets across all hosts
+const MAX_TOTAL_SOCKETS = 25;
+
+// Total number of sockets per each host
+const MAX_SOCKETS_PER_HOST = 5;
```
joe-yeager, line 14: This effectively limits the max number of of inflight API requests to a particular domain to 5.  If 5 requests are in flight, they will queue.  This value was just the value I found in my testing that seemed to greatly reduce the `ENOMEM` errors.  Instead of roughly 5%-10% `hs project upload` commands failing, this got it down to under 1% for me (I made it to 250 uploads before getting a `ENOMEM`)

http/getAxiosConfig.ts
```diff
@@ -4,6 +4,26 @@ import { getHubSpotApiOrigin } from '../lib/urls';
 import { AxiosConfigOptions } from '../types/Http';
 import { CLIConfig } from '../types/Config';
 import { AxiosRequestConfig } from 'axios';
+import https from 'https';
+import http from 'https';
+
+// Total number of sockets across all hosts
+const MAX_TOTAL_SOCKETS = 25;
```
joe-yeager, line 11: We could probably be more conservative with this number since this means there could be 25 idle socket connections in the worst case scenario.  The vast majority of our API calls seem to be to hubspot domains, so the likely hood of that happening feels slim.  Still, curious what others think this.

types/Utils.ts
camden11: Unfortunately I don't think we ever figured this out. @joe-yeager For context this particular line seems to format differently for different people and we haven't been able to figure out why

http/getAxiosConfig.ts
```diff
@@ -4,6 +4,26 @@ import { getHubSpotApiOrigin } from '../lib/urls';
 import { AxiosConfigOptions } from '../types/Http';
 import { CLIConfig } from '../types/Config';
 import { AxiosRequestConfig } from 'axios';
+import https from 'https';
+import http from 'https';
```
camden11, line 8: should this be `http`?

joe-yeager: It sure should be, great catch.

types/Utils.ts
joe-yeager: @brandenrodgers @camden11 @kemmerle 
How do you all feel about adding a prettier ignore block around this?  It's my least favorite solution in these cases, but it feels better than committing with the `-n` flag

camden11: Fine by me, definitely better than it going back and forth. Feel free to include that as part of this PR

camden11 (approved):
LGTM!

http/getAxiosConfig.ts
joe-yeager: Commit to address the issue and update tests to catch it https://github.com/HubSpot/hubspot-local-dev-lib/pull/139/commits/e1e396dee2a37247a204f04c3cd373df22ad9df0

PR #140, opened 2024-05-29
Fix issue with downloading images from GitHub repos

kemmerle:
## Description and Context
We discovered a bug with the `hs project create` command: image files were not being downloaded correctly. Images were being stored as strings and written in `utf8`. Now we store them as array buffers and write them out to a binary file. 

## TODO

- [x] Test out the PR. You can download a sample template with png file with the `hs project create --templateSource=kemmerle/sample-projects` command. 
- [x] Address any feedback

## Who to Notify

@sejal27 @brandenrodgers @camden11 @joe-yeager

### DIFF
--- a/lib/__tests__/github.ts
+++ b/lib/__tests__/github.ts
@@ -1,17 +1,21 @@
+import fs from 'fs-extra';
 import {
   fetchReleaseData,
   cloneGithubRepo,
   listGithubRepoContents,
   fetchFileFromRepository,
+  fetchGitHubRepoContentFromDownloadUrl,
 } from '../github';
 import {
   fetchRepoFile as __fetchRepoFile,
   fetchRepoReleaseData as __fetchRepoReleaseData,
   fetchRepoAsZip as __fetchRepoAsZip,
   fetchRepoContents as __fetchRepoContents,
+  fetchRepoFileByDownloadUrl as __fetchRepoFileByDownloadUrl,
 } from '../../api/github';
 import { extractZipArchive as __extractZipArchive } from '../archive';
 
+jest.mock('fs-extra');
 jest.mock('../../api/github');
 jest.mock('../archive');
 
@@ -30,6 +34,10 @@ const fetchRepoContents = __fetchRepoContents as jest.MockedFunction<
 const extractZipArchive = __extractZipArchive as jest.MockedFunction<
   typeof __extractZipArchive
 >;
+const fetchRepoFileByDownloadUrl =
+  __fetchRepoFileByDownloadUrl as jest.MockedFunction<
+    typeof __fetchRepoFileByDownloadUrl
+  >;
 
 describe('lib/github', () => {
   describe('fetchFileFromRepository()', () => {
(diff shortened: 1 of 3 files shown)


lib/github.ts
```diff
@@ -148,11 +148,15 @@ async function fetchGitHubRepoContentFromDownloadUrl(
...
-      ? resp.data
-      : JSON.stringify(resp.data, null, 2);
-  fs.outputFileSync(dest, fileContents, 'utf8');
+  const contentType = resp.headers['content-type'];
+  let fileContents;
+
+  if (contentType.startsWith('text')) {
+    fileContents = Buffer.from(resp.data).toString('utf8');
+  } else {
+    fileContents = resp.data;
```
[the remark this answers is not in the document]
brandenrodgers: iirc there were some scenarios where the data needs to be stringified and some when it doesn't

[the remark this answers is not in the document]
kemmerle: In the `axios` call, I specified that the response type should always be `arrayBuffer`. But if the mime type begins with "text", then we extract the text string and encode it as 'utf8.' 

In the original code, if you received data whose mime type was not text, then you got a buffer instead of a string. Doing it in reverse is more reliable. It also works for both binary files (images) and text files.

lib/github.ts
```diff
@@ -148,11 +148,15 @@ async function fetchGitHubRepoContentFromDownloadUrl(
   downloadUrl: string
 ): Promise<void> {
   const resp = await fetchRepoFileByDownloadUrl(downloadUrl);
-  const fileContents =
-    typeof resp.data === 'string'
-      ? resp.data
-      : JSON.stringify(resp.data, null, 2);
-  fs.outputFileSync(dest, fileContents, 'utf8');
+  const contentType = resp.headers['content-type'];
```
joe-yeager, line 151: Could we add some tests to this method while we are here to make sure we preserve the behavior in the future?

joe-yeager (commented):
LGTM, just a question about adding some tests to the method while we are here.

lib/github.ts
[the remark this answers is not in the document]
brandenrodgers: Awesome, thanks for clarifying!

lib/__tests__/github.ts
```diff
@@ -48,6 +56,51 @@ describe('lib/github', () => {
...
+        headers: {
+          'content-type': 'application/octet-stream',
+        },
+        data: Buffer.from([0x01, 0x02, 0x03]),
+        status: 200,
+        statusText: 'OK',
+        config: {},
+      };
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
kemmerle, line 94: I'm having an issue with TypeScript here that I'd love some advice on:

TypeScript is basically requiring that we mock out everything in the `config` property, as so:

`config: {
  headers: {
    set: jest.fn(),
    etc. 
  }
}

Is that really necessary for this single test? It seems like overkill, so I disabled the error for now. Is there anything I'm missing?

kemmerle: ^^ Talked this over with Camden in Slack, and he agreed it's overkill to include all the data.

lib/github.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/140/commits/40ac634e64750d92576def4dd61b83471c676656

joe-yeager (approved):

camden11 (approved):

PR #144, opened 2024-06-04
Fix prettier

camden11:
## Description and Context
This PR _should_ fix all of the issues we've been having with prettier 🤞. There were a few things that I think were causing us some trouble
* We were using `prettier.config.js` rather than `.prettierrc`. I'm not positive about this, but I think that it's possible that some people could have had `.prettierrc` in higher up directories interfering with the one here, since `.prettierrc` takes priority over `prettier.config.js` (see [docs](https://prettier.io/docs/en/configuration)). Regardless, `.prettierrc` is the recommended way to do prettier config.
* The `prettier:write` and `lint` functions in `package.json` were attempting to check only js, ts, and JSON files, but seemingly were only working this on some people's machines. I updated these scripts to check the entire directory. If we ever need to only check certain file types again, we can use `.prettierignore`, but I didn't see a downside to running the check everywhere (it already ignores gitignored files)

## Testing
Make some changes (especially to problem files like `Utils.ts`) and confirm formatting on your machine is consistent with this PR. Make sure formatting within VScode matches formatting by running `prettier:write`

## Who to Notify


joe-yeager (approved):

brandenrodgers (approved):
lgtm! I pulled this and my computer is happy with it 👌

PR #146, opened 2024-06-04
chore: Set jest `clearMocks` to true.  Remove manual mock clear calls

joe-yeager:
## Description and Context

Jest sets [`clearMocks`](https://jestjs.io/docs/configuration#clearmocks-boolean) to false by default which means that unless mocks are explicitly cleared between test invocations, the mocks states are carried between tests. which can lead to having tests that aren't well isolated and depend on the tests that come before them.

## Who to Notify

@brandenrodgers @kemmerle @camden11

### DIFF
--- a/lib/__tests__/uploadFolder.ts
+++ b/lib/__tests__/uploadFolder.ts
@@ -85,7 +85,6 @@ describe('lib/cms/uploadFolder', () => {
     createIgnoreFilter.mockImplementation(() => () => true);
   });
   beforeEach(() => {
-    FieldsJs.mockClear();
     createTmpDirSync.mockReset();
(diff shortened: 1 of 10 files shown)


lib/__tests__/uploadFolder.ts
```diff
@@ -164,6 +163,7 @@ describe('lib/cms/uploadFolder', () => {
     });
 
     it('deletes the temporary directory', async () => {
+      createTmpDirSync.mockReturnValue('folder');
```
joe-yeager, line 166: The test before this was setting up this mock, so this started failing when `clearMocks` was set to true.

PR #148, opened 2024-06-10
chore: Further expand test coverage

joe-yeager:
## Description and Context

Expand the test coverage further

## Who to Notify

@brandenrodgers @kemmerle @camden11


lib/__tests__/notify.ts
```diff
@@ -0,0 +1,94 @@
...
+      // Advance all of the timers to trigger the debounce
+      await jest.runAllTimersAsync();
+
+      expect(throwErrorWithMessageMock).toHaveBeenCalledTimes(1);
+      expect(throwErrorWithMessageMock).toHaveBeenCalledWith(
+        'utils.notify.errors.filePath',
+        { filePath: filePathToNotify },
+        error
+      );
+    });
```
joe-yeager, line 92: Not crazy about this assertion, but it seems like the best that we can do because the exception doesn't happen until `jest.runAllTimersAsync` finishes and the `debounce` method triggers.

camden11 (approved):
You're really powering through these tests! LGTM

brandenrodgers:
No comments on the code in here (lgtm), but the test coverage in this repo is looking **_really_** nice! 😎

PR #149, opened 2024-06-14
Remove the meta.charset_guess type

brandenrodgers:
## Description and Context

The BE team is removing this field. We don't use it anywhere in our lib or in consuming packages so it's safe to remove.

Note: Prettier was angry with me about the notify.ts file and that's why there are changes in there. I'm on the latest main branch, so I'm not sure what's going on there.
## TODO

N/A
## Who to Notify

I'll ping the BE team internally to let them know about this

### DIFF
--- a/lib/__tests__/notify.ts
+++ b/lib/__tests__/notify.ts
@@ -50,7 +50,9 @@ describe('lib/notify', () => {
             resolve();
           })
         );
-        output = `${output}${new Date(now).toISOString()} ${actionType}: ${filePath}\n`;
+        output = `${output}${new Date(
+          now
+        ).toISOString()} ${actionType}: ${filePath}\n`;
       }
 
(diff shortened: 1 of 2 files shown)


lib/__tests__/notify.ts
```diff
@@ -50,7 +50,9 @@ describe('lib/notify', () => {
             resolve();
           })
         );
-        output = `${output}${new Date(now).toISOString()} ${actionType}: ${filePath}\n`;
+        output = `${output}${new Date(
```
joe-yeager, line 53: Uh oh, is prettier acting up again?

joe-yeager (approved):

lib/__tests__/notify.ts
brandenrodgers: I'm hoping this is just a one-off thing and not the sign of this acting up 🤞

PR #151, opened 2024-06-17
POC: Wrap AxiosErrors in custom error object

joe-yeager:
## Description and Context

I think our http methods could be considered a leaky abstraction because we are directly exposing `AxiosErrors` which requires the projects depending on these methods to write `axios` specific logic in their code base, such as checking `e.response && e.response.status` in catch blocks.  What I am proposing is that we globally wrap all `AxiosErrors` in our own custom `Error` to keep the implementation details limited to this repository.

## Who to Notify

@brandenrodgers @camden11 @kemmerle

### DIFF
--- a/http/index.ts
+++ b/http/index.ts
@@ -1,7 +1,12 @@
 import path from 'path';
 import fs from 'fs-extra';
 import contentDisposition from 'content-disposition';
-import axios, { AxiosRequestConfig, AxiosResponse, isAxiosError } from 'axios';
+import axios, {
+  AxiosRequestConfig,
+  AxiosResponse,
+  InternalAxiosRequestConfig,
+  isAxiosError,
+} from 'axios';
 
 import { getAccountConfig } from '../config';
 import { USER_AGENTS, getAxiosConfig } from './getAxiosConfig';
@@ -15,18 +20,34 @@ import { i18n } from '../utils/lang';
 
 const i18nKey = 'http.index';
 
-export class HubSpotHttpError extends Error {
-  private status?: number;
+export class HubSpotHttpError<T = unknown, D = unknown> extends Error {
+  public status?: number;
+  public response?: AxiosResponse<T, D>;
+  public config?: InternalAxiosRequestConfig<D>;
+  public code?: string;
+  public request?: unknown;
+  public isHubSpotHttpError = true;
+
   constructor(message?: string, options?: ErrorOptions) {
     super(message, options);
     this.name = 'HubSpotHttpError';
-    if (options && isAxiosError(options.cause)) {
-      this.status = options.cause.status;
+
+    if (options && isAxiosError(options.cause) && options.cause.response) {
+      this.status = options.cause.response.status;
+
+      // Add these for backwards compatibility until we have updated all the checks
+      // in the CLI for the Axios implementation details and then we can remove these or
+      // keep whatever we find useful in custom fields on this object
+      const { response, request, config, code } = options.cause;
+      this.response = response;
+      this.request = request;
+      this.config = config;
+      this.code = code;
     }
   }
 }
 
-axios.interceptors.response.use(undefined, function (error) {
+axios.interceptors.response.use(undefined, error => {
   // Wrap all axios errors in our own Error class
   return Promise.reject(new HubSpotHttpError(error.message, { cause: error }));
 });
--- a/http/index.ts
+++ b/http/index.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import fs from 'fs-extra';
 import contentDisposition from 'content-disposition';
-import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
+import axios, { AxiosRequestConfig, AxiosResponse, isAxiosError } from 'axios';
 
 import { getAccountConfig } from '../config';
 import { USER_AGENTS, getAxiosConfig } from './getAxiosConfig';
@@ -14,6 +14,22 @@ import { logger } from '../lib/logger';
 import { i18n } from '../utils/lang';
 
 const i18nKey = 'http.index';
+
+export class HubSpotHttpError extends Error {
+  private status?: number;
+  constructor(message?: string, options?: ErrorOptions) {
+    super(message, options);
+    this.name = 'HubSpotHttpError';
+    if (options && isAxiosError(options.cause)) {
+      this.status = options.cause.status;
+    }
+  }
+}
+
+axios.interceptors.response.use(undefined, function (error) {
+  // Wrap all axios errors in our own Error class
+  return Promise.reject(new HubSpotHttpError(error.message, { cause: error }));
+});
 
 export function addUserAgentHeader(key: string, value: string) {
   USER_AGENTS[key] = value;


camden11:
This is a really good idea - we should absolutely be giving consuming libraries everything they need to interface with the errors we throw here. Should definitely be very careful to make sure we don't lose any info contained in the errors along the way, but I don't really see a downside to doing this

joe-yeager:
> Should definitely be very careful to make sure we don't lose any info contained in the errors along the way
@camden11 💯 

My understanding is that by creating the new error type and setting the error thrown by axios as `cause` we preserve all of the stack trace context.  We can also still access the axios error fields in the event there is a piece of information we don't have available on the top level of the new error.

http/index.ts
```diff
@@ -15,6 +20,54 @@ import { i18n } from '../utils/lang';
 
 const i18nKey = 'http.index';
 
+export class HubSpotHttpError<T = unknown, D = unknown> extends Error {
```
camden11, line 23: Could see this making sense having its own file (theres a precedent with `HubspotAuthError`)

joe-yeager: Good call, I'll move it over into it's own thing.

models/HubSpotHttpError.ts
```diff
@@ -0,0 +1,46 @@
...
+      // in the CLI for the Axios implementation details, and then we can remove these or
+      // keep whatever we find useful in custom fields on this object
+      const { response, request, config, code } = options.cause;
+      this.response = response;
+      this.request = request;
+      this.config = config;
+
+      // Add any custom fields we feel are necessary or make our collective lives easier
+      this.code = code;
+      if (response) {
```
brandenrodgers, line 32: For debugging purposes should we include more of the response information here? Things like headers and config maybe ([response schema](https://axios-http.com/docs/res_schema))?

errors/apiErrors.ts
```diff
@@ -10,11 +10,38 @@ import { i18n } from '../utils/lang';
...
+  code?: string;
+} {
+  if (isHubSpotHttpError(error)) {
+    return {
+      data: error.data,
+      status: error.status,
+      code: error.code,
+    };
+  }
+  if (isAxiosError(error)) {
```
brandenrodgers, line 33: Can you clarify why we still need this line if all calls are using the new Axios interceptor that will convert the errors?

errors/apiErrors.ts
```diff
@@ -238,17 +272,17 @@ export function getAxiosErrorWithContext(
  * @throws
  */
 export function throwApiError(
-  error: AxiosError,
+  error: HubSpotHttpError,
   context: AxiosErrorContext = {}
 ): never {
   if (error.isAxiosError) {
```
brandenrodgers, line 278: I can't remember why we originally included this `isAxiosError` check. Do we still need it?

errors/apiErrors.ts
```diff
@@ -29,45 +56,53 @@ export function isSpecifiedError(
...
 
-export function isTimeoutError(err: Error | AxiosError): boolean {
+export function isTimeoutError(err: Error | HubSpotHttpError): boolean {
   return isSpecifiedError(err, { code: 'ETIMEDOUT' });
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function isApiUploadValidationError(err: AxiosError<any>): boolean {
+export function isApiUploadValidationError(err: HubSpotHttpError): boolean {
   return (
```
brandenrodgers, line 101: Could this util make use of `isSpecifiedError`?

errors/apiErrors.ts
```diff
@@ -29,45 +56,53 @@ export function isSpecifiedError(
...
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const error = (err && (err.cause as AxiosError<any>)) || err;
-  const statusCodeErr = !statusCode || error.response?.status === statusCode;
-  const categoryErr = !category || error.response?.data?.category === category;
-  const subCategoryErr =
-    !subCategory || error.response?.data?.subCategory === subCategory;
-  const errorTypeErr =
-    !errorType || error.response?.data?.errorType === errorType;
-  const codeError = !code || error.code === code;
+  if (!isHubSpotHttpError(err) && !isAxiosError(err.cause)) {
```
brandenrodgers, line 59: Is our call to `isAxiosError(err.cause)` for backwards compatibility?

joe-yeager: Yeah, it's for backwards compatibility.  I flagged this as ready to review and then started changing more things so it's kind of in between state so I'm going to flip it back to draft and make the changes to the CLI simultaneously since that is the main consumer of the this package.  If I do that we can ditch a lot of this logic that was added for backwards compatability.

errors/apiErrors.ts
joe-yeager: Probably not, I'm going to work on the integration in tandem with this and start to remove some of the backwards compatibility holdovers.

models/HubSpotHttpError.ts
joe-yeager: Good call, I'll add headers to the top level

errors/apiErrors.ts
joe-yeager: We do not need this anymore, I was leaving it around for backwards compatibility, but we can remove it and update the CLI simultaneously

errors/apiErrors.ts
joe-yeager: Yep, it can use it for the first part of the conditional check.  Updated

errors/apiErrors.ts
joe-yeager: Removed

errors/__tests__/apiErrors.ts
```diff
@@ -23,33 +24,32 @@ export const newError = (overrides = {}): BaseError => {
...
+      isAxiosError: true,
+      name: 'HubSpotHttpError',
+      response: {
+        request: {
+          href: 'http://example.com/',
+          method: 'GET',
+        },
+        data: {},
+        headers: {},
+        status: 200,
```
kemmerle, line 40: Nit: Do we want another 400/500 status code?

errors/__tests__/apiErrors.ts
```diff
@@ -80,77 +80,60 @@ describe('errors/apiErrors', () => {
...
+      const error1 = newHubSpotHttpError({
         response: { status: 403, data: { category: 'MISSING_SCOPES' } },
       });
       expect(isMissingScopeError(error1)).toBe(true);
     });
 
     it('returns false for non missing scope errors', () => {
-      const error1 = newAxiosError({
+      const error1 = newHubSpotHttpError({
         response: { status: 400, data: { category: 'MISSING_SCOPES' } },
```
kemmerle, line 94: Do we need to rewrite this test with another error category? Isn't this the test for non-missing scope errors?

kemmerle: I see the same pattern below, am I missing something?

kemmerle (approved):
Besides the one tiny nit, this all looks great to me. Particularly like how you made the request fields top level in HubSpot errors. Chef's kiss👩‍🍳 💋

errors/apiErrors.ts
```diff
@@ -116,16 +112,15 @@ export function parseValidationErrors(
...
   const validationErrorMessages = parseValidationErrors(error?.response?.data);
   if (validationErrorMessages.length) {
     return new Error(validationErrorMessages.join(' '), { cause: error });
   }
 }
 
-export function getAxiosErrorWithContext(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  error: AxiosError<any>,
+export function getHubSpotHttpErrorWithContext(
```
brandenrodgers, line 122: Just recording here that this will be a major version bump since it's a breaking change (totally fine, just wanted to add a comment for posterity)

errors/__tests__/apiErrors.ts
joe-yeager: Good eye, this was carryover, but I agree it feels weird for an error to be 200.  All of the tests appear to be overriding it to test the specific cases, so I removed the status from this object.

PR #152, opened 2024-06-18
Add private app user token manger for use with local dev server

wung-hubspot:
## Description and Context
Private app user tokens are temporary Private App tokens scoped to the portal and user of the Personal Access Key. Using this user can avoid grabbing the real Private app token that is used for in the serverless instances. 

Adds api calls to private app user token endpoints in localdevauth and creates a manager class to manage these token refreshes when they expire. 

## Screenshots

Screenshot 2024-06-18 at 11 17 12 AM
local cli run with logs showing usage.

### DIFF
--- a/api/privateAppUserToken.ts
+++ b/api/privateAppUserToken.ts
@@ -25,13 +25,13 @@ export async function fetchPrivateAppUserToken(
 export async function createPrivateAppUserToken(
   accountId: number,
   appId: number,
-  scopeGroups: Array<string> = [],
+  scopeGroups?: Array<string>,
   expiresAt?: string
 ): Promise<PrivateAppUserTokenResponse> {
   return http.post(accountId, {
     url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
     data: {
-      ...(scopeGroups.length && { scopeGroups: scopeGroups }),
+      ...(scopeGroups && { scopeGroups: scopeGroups }),
       ...(expiresAt && { expiresAt: expiresAt }),
     },
   });
@@ -41,14 +41,14 @@ export async function updatePrivateAppUserToken(
   accountId: number,
   appId: number,
   userTokenKey: string,
-  scopeGroups: Array<string> = [],
+  scopeGroups?: Array<string>,
   expiresAt?: string
 ): Promise<PrivateAppUserTokenResponse> {
   return http.put(accountId, {
     url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
     data: {
       privateAppUserTokenKey: userTokenKey,
-      ...(scopeGroups.length && { scopeGroups: scopeGroups }),
+      ...(scopeGroups && { scopeGroups: scopeGroups }),
       ...(expiresAt && { expiresAt: expiresAt }),
     },
   });
--- a/api/privateAppUserToken.ts
+++ b/api/privateAppUserToken.ts
@@ -1,0 +1,55 @@
+import http from '../http';
+
+const LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH =
+  'localdevauth/v1/private-app/user-token';
+
+export type PrivateAppUserTokenResponse = {
+  userId: number;
+  portalId: number;
+  appId: number;
+  scopeGroups: Array<string>;
+  userTokenKey: string;
+  expiresAt: string;
+  clientId: string;
+};
+
+export async function fetchPrivateAppUserToken(
+  accountId: number,
+  appId: number
+): Promise<PrivateAppUserTokenResponse> {
+  return http.get(accountId, {
+    url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
+  });
+}
+
+export async function createPrivateAppUserToken(
+  accountId: number,
+  appId: number,
+  scopeGroups: Array<string> = [],
+  expiresAt?: string
+): Promise<PrivateAppUserTokenResponse> {
+  return http.post(accountId, {
+    url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
+    data: {
+      ...(scopeGroups.length && { scopeGroups: scopeGroups }),
+      ...(expiresAt && { expiresAt: expiresAt }),
+    },
+  });
+}
+
+export async function updatePrivateAppUserToken(
+  accountId: number,
+  appId: number,
+  userTokenKey: string,
+  scopeGroups: Array<string> = [],
+  expiresAt?: string
+): Promise<PrivateAppUserTokenResponse> {
+  return http.put(accountId, {
+    url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
+    data: {
+      privateAppUserTokenKey: userTokenKey,
+      ...(scopeGroups.length && { scopeGroups: scopeGroups }),
+      ...(expiresAt && { expiresAt: expiresAt }),
+    },
+  });
+}
--- a/lang/en.json
+++ b/lang/en.json
@@ -238,7 +238,8 @@
       "refreshScheduled": "Scheduling refresh of token for appId~{{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
       "errors": {
         "noScopes": "Unable to use private app user tokens for accountId~{{ accountId }}. Please regenerate the Personal Access Key.",
-        "apiError": "Unable to get or create private app user token for appId~{{ appId }} on accountId~{{ accountId }}. {{ messageDetail }} failed."
+        "apiError": "Unable to get or create private app user token for appId~{{ appId }} on accountId~{{ accountId }}. {{ messageDetail }} failed.",
+        "refreshFailed": "Unable to refresh private app user token for appId~{{ appId }} on accountId~{{ accountId }}."
       }
     }
   },
--- a/lang/en.json
+++ b/lang/en.json
@@ -230,16 +230,16 @@
       "standard": "standard"
     },
     "PrivateAppUserTokenManager": {
-      "enabled": "Private app user tokens is enabled for accountId~{{ accountId }}",
-      "disabled": "Private app user tokens is not enabled for accountId~{{ accountId }}, skipping call to fetch for appId~{{ appId }}",
-      "create": "Creating new private app user token for appId~{{ appId }}",
-      "cached": "Using cached private app user token for appId~{{ appId }}",
-      "refresh": "Updating or rereshing private app user token for appId~{{ appId }} ",
-      "refreshScheduled": "Scheduling refresh of token for appId~{{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
+      "enabled": "Private app user tokens is enabled for accountId {{ accountId }}",
+      "disabled": "Private app user tokens is not enabled for accountId {{ accountId }}, skipping call to fetch for appId {{ appId }}",
+      "create": "Creating new private app user token for appId {{ appId }}",
+      "cached": "Using cached private app user token for appId {{ appId }}",
+      "refresh": "Updating or rereshing private app user token for appId {{ appId }} ",
+      "refreshScheduled": "Scheduling refresh of token for appId {{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
       "errors": {
-        "noScopes": "Unable to use private app user tokens for accountId~{{ accountId }}. Please regenerate the Personal Access Key.",
-        "apiError": "Unable to get or create private app user token for appId~{{ appId }} on accountId~{{ accountId }}. {{ messageDetail }} failed.",
-        "refreshFailed": "Unable to refresh private app user token for appId~{{ appId }} on accountId~{{ accountId }}."
+        "noScopes": "Unable to use private app user token for appId {{ appId }} since updated App function scope is missing on accountId {{ accountId }}. To add App function scope to that account, deactivate the existing Personal Access Key, and generate a new one that includes the App functions scope. Then run hs auth and reauthenticate accountId {{ accountId }}.",
+        "apiError": "Unable to get or create private app user token for appId {{ appId }} on accountId {{ accountId }}. {{ messageDetail }}",
+        "refreshFailed": "Unable to refresh private app user token for appId {{ appId }} on accountId {{ accountId }}."
       }
     }
   },
--- a/lang/en.json
+++ b/lang/en.json
@@ -347,6 +347,15 @@
         "409": "Failed to assign port. Server with instanceId {{ instanceId }} is already running on port {{ port }}",
         "400": "Invalid port requested. Port must be between {{ minPort }} and {{ maxPort }}."
       }
+    },
+    "PrivateAppUserTokenManager": {
+      "disabled": "Private application user tokens is not enabled for accountId~{{ accountId }}, skipping call to fetch for appId~{{ appId }}",
+      "create": "Creating new private applications user token for appId~{{ appId }}",
+      "refresh": "Updating or rereshing private application user token for appId~{{ appId }} ",
+      "refreshScheduled": "Scheduling refresh of token for appId~{{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
+      "errors": {
+        "noScopes": "Unable to use temporary Private Application User tokens for accountId~{{ accountId }}. Please regenerate the Personal Access Key."
+      }
     }
   },
   "http": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -228,6 +228,18 @@
       "developerTest": "test account",
       "appDeveloper": "dev account",
       "standard": "standard"
+    },
+    "PrivateAppUserTokenManager": {
+      "enabled": "Private application user tokens is enabled for accountId~{{ accountId }}",
+      "disabled": "Private application user tokens is not enabled for accountId~{{ accountId }}, skipping call to fetch for appId~{{ appId }}",
+      "create": "Creating new private applications user token for appId~{{ appId }}",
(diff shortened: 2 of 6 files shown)


lib/personalAccessKey.ts
```diff
@@ -162,22 +179,16 @@ export async function accessTokenForPersonalAccessKey(
...
-    accountId,
-    personalAccessKey,
-    authTokenInfo && authTokenInfo.expiresAt,
-    env
-  );
+  const accessTokenResponse = await getAccessTokenResponse(accountId);
   return accessTokenResponse?.enabledFeatures;
 }
 
+export async function scopeGroupsForPersonalAccessKey(
```
joe-yeager, line 186: Nit: We call it a `AccessToken` in `getAccessTokenResponse` and `AccessKey` here, could we standardize

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  private async getExistingToken(
+    appId: number
+  ): Promise<PrivateAppUserTokenResponse | null> {
+    try {
+      return await fetchPrivateAppUserToken(this.accountId, appId);
+    } catch (err) {
+      if (err instanceof AxiosError) {
+        const axiosError = err as AxiosError;
```
joe-yeager, line 213: If we use `if(isAxiosError(err))` it **_should_** type narrow and we can ditch the type assertion `err as AxiosError`

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+        );
+        return this.tokenMap.get(appId)!.userTokenKey;
+      } else {
+        const token = await this.createOrGetActiveToken(appId, scopeGroups);
+        this.setCacheAndRefresh(appId, token, scopeGroups);
+        return token.userTokenKey;
+      }
+    } catch (err) {
+      let messageDetail = 'Unknown error';
+      if (err instanceof AxiosError) {
```
joe-yeager, line 94: We could also use `isAxiosError` here

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+        logger.error(err as AxiosError);
+        messageDetail = getAxiosErrorWithContext(err as AxiosError).message;
+      } else if (err instanceof Error) {
+        logger.error(err as Error);
```
joe-yeager, line 98: I think just extracting the message on the line below is probably good enough, this stack trace could be rather large and most of the context is probably not relevant to the end user.

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+        return this.tokenMap.get(appId)!.userTokenKey;
+      } else {
+        const token = await this.createOrGetActiveToken(appId, scopeGroups);
+        this.setCacheAndRefresh(appId, token, scopeGroups);
+        return token.userTokenKey;
+      }
+    } catch (err) {
+      let messageDetail = 'Unknown error';
+      if (err instanceof AxiosError) {
+        logger.error(err as AxiosError);
```
joe-yeager, line 95: `AxiosErrors` are pretty large, maybe just extracting the message like you do below is sufficient

lang/en.json
```diff
@@ -228,6 +228,19 @@
...
+      "enabled": "Private app user tokens is enabled for accountId~{{ accountId }}",
+      "disabled": "Private app user tokens is not enabled for accountId~{{ accountId }}, skipping call to fetch for appId~{{ appId }}",
+      "create": "Creating new private app user token for appId~{{ appId }}",
+      "cached": "Using cached private app user token for appId~{{ appId }}",
+      "refresh": "Updating or rereshing private app user token for appId~{{ appId }} ",
+      "refreshScheduled": "Scheduling refresh of token for appId~{{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
+      "errors": {
+        "noScopes": "Unable to use private app user tokens for accountId~{{ accountId }}. Please regenerate the Personal Access Key.",
+        "apiError": "Unable to get or create private app user token for appId~{{ appId }} on accountId~{{ accountId }}. {{ messageDetail }} failed.",
+        "refreshFailed": "Unable to refresh private app user token for appId~{{ appId }} on accountId~{{ accountId }}."
```
camden11, line 242: You can remove the `~`s here and just say `appId {{ appId }}`, we don't use notation like that anywhere else

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  async init(): Promise<void> {
+    const scopeGroups = new Set<string>(
+      await scopeGroupsForPersonalAccessKey(this.accountId)
+    );
+    if (
+      !scopeGroups.has(USER_TOKEN_READ) ||
+      !scopeGroups.has(USER_TOKEN_WRITE)
+    ) {
+      logger.warn(
```
camden11, line 39: We try to use `logger.log` and `logger.warn` as little as possible in `local-dev-lib` in favor of the CLI handling logs that users actually see. This looks like a good place to use `throwErrorWithMessage` to throw this error message and catch and log it on the CLI side

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+        messageDetail = err.message;
+      }
+      logger.warn(
+        i18n(`${i18nKey}.errors.apiError`, {
+          accountId: this.accountId,
+          appId: appId,
+          messageDetail: messageDetail,
+        })
+      );
+    }
```
camden11, line 108: Same as above, the error logging is best handled on the CLI. You can use `throwErrorWithMessage` to let this class have control over the actual message, and include the original error as the `cause` arg.

You can also use `error.isAxiosError` rather than instance of

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    return this.enabled;
+  }
+
+  private setCacheAndRefresh(
+    appId: number,
+    token: PrivateAppUserTokenResponse,
+    scopeGroups: string[]
+  ) {
+    if (token === undefined || token === null) {
+      logger.warn(
```
camden11, line 121: Same here re: throwing errors

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    );
+  }
+
+  private async getExistingToken(
+    appId: number
+  ): Promise<PrivateAppUserTokenResponse | null> {
+    try {
+      return await fetchPrivateAppUserToken(this.accountId, appId);
+    } catch (err) {
+      if (err instanceof AxiosError) {
```
camden11, line 212: same here re `isAxiosError`

joe-yeager (commented):
I took a first pass, I can dig deeper later today

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    const newToken = await updatePrivateAppUserToken(
+      this.accountId,
+      appId,
+      userTokenKey,
+      scopeGroups
+    );
+    if (newToken) {
+      this.setCacheAndRefresh(appId, newToken, scopeGroups);
+    } else {
+      logger.warn(
```
camden11, line 241: Same re: throwing errors

lib/personalAccessKey.ts
```diff
@@ -133,6 +133,23 @@ async function getNewAccessToken(
   return accessTokenResponse;
 }
 
+async function getAccessTokenResponse(accountId: number) {
```
camden11, line 136: Missing return type here

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+
+  cleanup() {
+    this.tokenMapIntervalId.forEach(timeoutId => clearInterval(timeoutId));
+    this.tokenMapIntervalId.clear();
+    this.tokenMap.clear();
+  }
+
+  async getPrivateAppToken(
```
brandenrodgers, line 60: Have we landed on an official name for these yet? They are different from standard private App tokens right?

api/privateAppUserToken.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function createPrivateAppUserToken(
+  accountId: number,
+  appId: number,
+  scopeGroups: Array<string> = [],
+  expiresAt?: string
+): Promise<PrivateAppUserTokenResponse> {
+  return http.post(accountId, {
+    url: `${LOCALDEVAUTH_API_PRIVATE_APP_USER_TOKEN_PATH}/${appId}`,
+    data: {
+      ...(scopeGroups.length && { scopeGroups: scopeGroups }),
```
brandenrodgers, line 34: Does this `createPrivateAppUserToken()` function support `scopeGroups` just for future use cases? From what I can tell nothing currently passes any scope groups to this. Could this be switched to an optional argument? Seems like it would be fine for it to just be undefined.

Somewhat related, another formatting option here that might be a little more straightforward is `scopeGroups: scopeGroups.length ? scopeGroups : undefined`.

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,249 @@
...
+          accountId: this.accountId,
+          appId: appId,
+        })
+      );
+      return;
+    }
+    try {
+      if (
+        this.tokenMap.has(appId) &&
+        PrivateAppUserTokenManager.doesTokenHaveAllScopes(
```
brandenrodgers, line 76: Could this check be switched to:
```js
if (this.doesTokenHaveAllScopes(appId, scopeGroups) { ... }
```

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+    this.tokenMapIntervalId.forEach(timeoutId => clearInterval(timeoutId));
+    this.tokenMapIntervalId.clear();
+    this.tokenMap.clear();
+  }
+
+  async getPrivateAppToken(
+    appId: number,
+    scopeGroups: string[] = []
+  ): Promise<string | undefined> {
+    if (!this.enabled) {
```
brandenrodgers, line 64: Should this throw? I'm assuming we wouldn't want the user to be able to continue if we hit this case for some reason

cc/ @camden11

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+  isEnabled(): boolean {
+    return this.enabled;
+  }
+
+  private setCacheAndRefresh(
+    appId: number,
+    token: PrivateAppUserTokenResponse,
+    scopeGroups: string[]
+  ) {
+    if (token === undefined || token === null) {
```
brandenrodgers, line 120: `!token` may be enough here, unless it can be zero

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+    );
+  }
+
+  private async getExistingToken(
+    appId: number
+  ): Promise<PrivateAppUserTokenResponse | null> {
+    try {
+      return await fetchPrivateAppUserToken(this.accountId, appId);
+    } catch (err) {
+      if (isAxiosError(err)) {
```
brandenrodgers, line 209: We have a util in this library that makes it easier to check for specific error types. It's called `isSpecifiedError()`

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+    const newToken = await updatePrivateAppUserToken(
+      this.accountId,
+      appId,
+      userTokenKey,
+      scopeGroups
+    );
+    if (newToken) {
+      this.setCacheAndRefresh(appId, newToken, scopeGroups);
+    } else {
+      throwErrorWithMessage(`${i18nKey}.errors.refreshFailed`, {
```
brandenrodgers, line 237: How does this thrown error impact the running process? Will it kill the local dev command? This would prevent future refreshes. I wonder if we should consider adding in a retry here since a failure would be pretty bad. I think we'd at least want to inform the user that the temporary PAT isn't working anymore.

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+      ) {
+        logger.debug(
+          i18n(`${i18nKey}.cached`, {
+            appId: appId,
+          })
+        );
+        return this.tokenMap.get(appId)!.userTokenKey;
+      } else {
+        const token = await this.createOrGetActiveToken(appId, scopeGroups);
+        this.setCacheAndRefresh(appId, token, scopeGroups);
```
brandenrodgers, line 89: `token` includes `token.scopeGroups` right? I'm thinking that we might be able to cut down on args in some places by accessing `token.scopeGroups` rather than passing around this `scopeGroups` variable. `setCacheAndRefresh()` for example, is only ever called when it's passed a new token.

lib/PrivateAppUserTokenManager.ts
camden11: Yeah, I guess it ultimately depends on how this returning `undefined` is handled in the CLI, but might make more sense to turn this debug into an error

lib/PrivateAppUserTokenManager.ts
```diff
@@ -0,0 +1,243 @@
...
+    token: PrivateAppUserTokenResponse,
+    scopeGroups: string[]
+  ) {
+    if (token === undefined || token === null) {
+      throwErrorWithMessage(`${i18nKey}.errors.refreshFailed`, {
+        accountId: this.accountId,
+        appId: appId,
+      });
+    }
+    this.tokenMap.set(appId, token);
```
brandenrodgers, line 126: Just riffing, but I wonder if there's a way to avoid needing to have both `tokenMap` and also `tokenMapInterval`.

For example:
```js
this.tokenMap[appId] = {
  value: token,
  refresh: setInterval(() => this.refresh(appId, token), interval)
}
```

lib/PrivateAppUserTokenManager.ts
wung-hubspot: If it's not enabled, it would function like it does now. The localdev server could work if the PAT is within their env file.

lib/PrivateAppUserTokenManager.ts
wung-hubspot: update, can't just pass the appId since this method is used elsewhere when fetching the token from the api

lib/PrivateAppUserTokenManager.ts
wung-hubspot: updated to just one map

lib/PrivateAppUserTokenManager.ts
wung-hubspot: left the variable for now; if we can grab the scope groups from the app.json file and pass it into the fetch it'll make sure that the token always has those scopes. The token's scope and the app json might be mismatched, though I can just delete all the SG stuff for now if we don't think we'll need it some point later

lang/en.json
```diff
@@ -228,6 +228,19 @@
       "developerTest": "test account",
       "appDeveloper": "dev account",
       "standard": "standard"
+    },
+    "PrivateAppUserTokenManager": {
+      "enabled": "Private app user tokens is enabled for accountId {{ accountId }}",
```
kemmerle, line 233: Should the verb here be plural? 

`Private app user tokens are enabled...`

lang/en.json
```diff
@@ -228,6 +228,19 @@
...
+      "disabled": "Private app user tokens is not enabled for accountId {{ accountId }}, skipping call to fetch for appId {{ appId }}",
+      "create": "Creating new private app user token for appId {{ appId }}",
+      "cached": "Using cached private app user token for appId {{ appId }}",
+      "refresh": "Updating or rereshing private app user token for appId {{ appId }} ",
```
kemmerle, line 237: rereshing -> refreshing typo

lang/en.json
```diff
@@ -228,6 +228,19 @@
...
+      "refreshScheduled": "Scheduling refresh of token for appId {{ appId }} expiring at {{ expiresAt }} to run at {{ refreshTime }} ",
+      "errors": {
+        "noScopes": "Unable to use private app user token for appId {{ appId }} since updated App function scope is missing on accountId {{ accountId }}. To add App function scope to that account, deactivate the existing Personal Access Key, and generate a new one that includes the App functions scope. Then run hs auth and reauthenticate accountId {{ accountId }}.",
```
kemmerle, line 240: You can use `uiCommandReference` for `hs auth`

lang/en.json
```diff
@@ -228,6 +228,19 @@
...
+        "apiError": "Unable to get or create private app user token for appId {{ appId }} on accountId {{ accountId }}. {{ messageDetail }}",
+        "refreshFailed": "Unable to refresh private app user token for appId {{ appId }} on accountId {{ accountId }}."
```
kemmerle, line 242: I'd add some instructions here for the user. Should they stop the server and reboot it?

lib/PrivateAppUserTokenManager.ts
kemmerle: See my comment above on the en.lyaml file. Should we give the user some instructions to troubleshoot?

lib/personalAccessKey.ts
```diff
@@ -133,6 +133,25 @@ async function getNewAccessToken(
...
 
+async function getNewAccessTokenByAccountId(
+  accountId: number
+): Promise<AccessToken> {
+  const account = getAccountConfig(accountId) as PersonalAccessKeyAccount;
+  if (!account) {
+    throwErrorWithMessage(`${i18nKey}.errors.accountNotFound`, { accountId });
+  }
+  const { auth, personalAccessKey, env } = account;
+  const authTokenInfo = auth && auth.tokenInfo;
```
kemmerle, line 144: What happens if `auth.tokenInfo` doesn't exist? Should we abort the call to `getNewAccessToken`?

lib/personalAccessKey.ts
kemmerle: @wung-hubspot, did you address this?

wung-hubspot:
Moving code

PR #155, opened 2024-06-21
chore: Make error handlers accept unknown and clean up type assertions

joe-yeager:
## Description and Context

**_This PR looks huge, but the big changes are localized to a couple of files that I tagged with comments, the rest are typescript changes, or updating tests._**

Updates functions that take errors as arguments to take type `unknown` and to type narrow the error to prevent the methods to require type assertions every time they are called.  

The reasoning behind this:
- The default type of the `error` in catch blocks is `unknown` and it is considered a bad practice to set it to anything other than `unknown`.  
- It required type assertions on every invocation and they are considered a bad practice because they are disabling the type system by us asserting we know the type with full certainty.  Because of this they give us a false sense of type security
 

## TODO

- [ ] Integrate this with the CLI and test things are working as expected

## Who to Notify

@camden11 @brandenrodgers @kemmerle

### DIFF
--- a/errors/apiErrors.ts
+++ b/errors/apiErrors.ts
@@ -1,20 +1,18 @@
-import { AxiosError } from 'axios';
-import {
-  GenericError,
-  AxiosErrorContext,
-  BaseError,
-  ValidationError,
-} from '../types/Error';
+import { AxiosError, isAxiosError } from 'axios';
+import { AxiosErrorContext, BaseError, ValidationError } from '../types/Error';
 import { HTTP_METHOD_VERBS, HTTP_METHOD_PREPOSITIONS } from '../constants/api';
 import { i18n } from '../utils/lang';
 import { throwError } from './standardErrors';
-import { HubSpotAuthError } from '../models/HubSpotAuthError';
+import {
+  HubSpotAuthError,
+  isHubSpotAuthError,
+} from '../models/HubSpotAuthError';
 import { HttpMethod } from '../types/Api';
 
 const i18nKey = 'errors.apiErrors';
 
 export function isSpecifiedError(
-  err: Error | AxiosError,
+  err: unknown,
   {
     statusCode,
     category,
@@ -29,6 +27,10 @@ export function isSpecifiedError(
     code?: string;
   }
 ): boolean {
+  if (!(err instanceof Error)) {
+    return false;
+  }
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const error = (err && (err.cause as AxiosError<any>)) || err;
   const statusCodeErr = !statusCode || error.response?.status === statusCode;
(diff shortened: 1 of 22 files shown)


errors/apiErrors.ts
```diff
@@ -1,20 +1,18 @@
-import { AxiosError } from 'axios';
-import {
-  GenericError,
-  AxiosErrorContext,
-  BaseError,
-  ValidationError,
-} from '../types/Error';
+import { AxiosError, isAxiosError } from 'axios';
```
joe-yeager, line 1: This is one of the files of interest

errors/standardErrors.ts
```diff
@@ -3,21 +3,28 @@ import { i18n } from '../utils/lang';
 
 import { BaseError } from '../types/Error';
 import { LangKey } from '../types/Lang';
-import { AxiosError } from 'axios';
 
-export function isSystemError(err: BaseError): boolean {
-  return err.errno != null && err.code != null && err.syscall != null;
+export function isSystemError(err: unknown): err is BaseError {
```
joe-yeager, line 7: This is one of the files of interest

errors/apiErrors.ts
```diff
@@ -49,41 +51,39 @@ export function isSpecifiedError(
...
     !!(err.response?.data?.message || !!err.response?.data?.errors)
   );
 }
 
 export function isSpecifiedHubSpotAuthError(
-  err: GenericError,
+  err: unknown,
   { status, category, subCategory }: Partial<HubSpotAuthError>
-): boolean {
+): err is HubSpotAuthError {
```
camden11, line 79: Wow I did not know you could do this. Huge.

errors/standardErrors.ts
```diff
@@ -3,21 +3,28 @@ import { i18n } from '../utils/lang';
...
+    err.errno != null &&
+    'code' in err &&
+    err.code != null &&
+    'syscall' in err &&
+    err.syscall != null
+  );
 }
 
-export function isFatalError(err: BaseError): boolean {
+export function isFatalError(err: unknown): err is BaseError {
```
camden11, line 19: Should this be `err is HubspotAuthError`?

camden11: Also realizing that this is redundant with `isHubSpotAuthError`, but I guess we can't really remove it without doing a major version bump

camden11 (approved):
This all looks great. This may have been obvious, but we were all fairly new to Typescript when we started working on this repo, and we never really had a good idea of how to handle typing the errors. This is a pretty big improvement I think.

One minor comment, but looks good to me.

errors/apiErrors.ts
joe-yeager: Yeah, they are called [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) and they are very powerful!

errors/standardErrors.ts
joe-yeager: Great question.  My reasoning for going `BaseError` over `HubSpotAuthError` was that this method might expand to cover other types of errors so I wanted to keep it generic.

models/HubSpotAuthError.ts
```diff
@@ -23,3 +23,7 @@ export class HubSpotAuthError extends Error {
       undefined;
   }
 }
+
+export function isHubSpotAuthError(err: unknown): err is HubSpotAuthError {
```
brandenrodgers, line 27: Do we need this util? It looks like we're only using it in one place. If we do, I think it might be better off living in another file rather than in the models/ folder

errors/standardErrors.ts
brandenrodgers: Maybe we can remove `isFatalError` in one of your follow up PR's? I'm guessing this rework of the error handling will eventually require a major version bump anyways.

models/HubSpotAuthError.ts
joe-yeager: We can ditch it, I was using it in more places initially.

errors/standardErrors.ts
joe-yeager: > Maybe we can remove isFatalError in one of your follow up PR's?

Yep, I will do this in the batch of PRs after the [AxiosError](https://github.com/HubSpot/hubspot-local-dev-lib/pull/151/files) rework

PR #156, opened 2024-07-10
add support for react modules to enable lang translations internally

jontallboy:
## Description and Context

Supports an internal use case where we need to translate lang keys for react modules. The changes in here support passing along JSON fields instead of a file path when it comes to react modules. 

## Screenshots

These changes, along with some changes to an internal lib enable functionality like you will see below for generating lang keys on internal HS projects:
translate-react-fields

### DIFF
--- a/lib/cms/handleFieldsJS.ts
+++ b/lib/cms/handleFieldsJS.ts
@@ -116,7 +116,6 @@ export class FieldsJs {
 
   // Create a temporary file with the fields data
   private createTempFile(data: JSON): Promise<string> {
-    console.log(data);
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):
The changes in here lgtm 👍 (I haven't tested them locally though). From what I can tell, this is backwards compatible, right?

jontallboy:
@brandenrodgers, correct this is backward compat and CAT has tested with success, but there are a few discussions happening around this on our side related to the internal lang use case, so I am going to hold off on merging for now. Thanks for giving this a review!

PR #158, opened 2024-07-17
Allow different types of for Data for axios request config

bmatto:
## Description and Context

Axios allows for more than just a plain JSON object for `data` in the config - see https://axios-http.com/docs/req_config. We have a need to pass a string here, and the PR just opens up the type based on the documentation.


camden11 (approved):
LGTM! Thanks for providing that docs link

PR #159, opened 2024-07-18
feat!: Refactor error handling

joe-yeager:
## Description and Context
_**Note: This will be a breaking change, so this PR is targeting the next branch and will be in the next major**_

#### Removals
- Removed all thrower methods because they were thin wrappers obscure our intent by adding an additional layer in front of the throw.  Explicitly throwing the error makes our intentions clearer, and actually makes our code easier to statically analyze.  The change uncovered 2 cases where we are calling `throwX` in a try block, which then obscures the intended error message with a new more generic error message.  Methods removed:
  - `throwError` 
  - `throwErrorWithMessage`   
  - `throwApiError`
  - `throwApiUploadError`
  - `throwValidationError`
  - `throwAuthErrorWithMessage`
  - `throwFileSystemError`
- Removes `HubSpotAuthError`, because we now have `HubSpotHttpError` and we can check if it is an auth error using `isAuthError` in the `errors` module
- `parseValidationErrors` is no longer exported
- `getHubSpotHttpErrorWithContext` is no longer exported

#### Additions
- Creates a new `Error` object `FileSystemError` to replace the need for `throwFileSystemError`

#### Changes
- Compresses `apiErrors`, `standardErrors`, and `fileSystemErrors` into a single `errors` module.
- Moves all of the logic for extraction of error messages from http responses into the `HubSpotHttpError` class and makes it an automatic operation.  
  - This means the `HubSpotHttpError` with have the user friendly http error message in it's `message` property automatically.
  - The extraction of the `context` object is automatic on instantiation of the `HubSpotHttpError`, but can be overridden

## TODO

- [x] Finish unit testing
- [x] FInish integration with CLI: https://github.com/HubSpot/hubspot-cli/pull/1118
- [x] Update errors/README.md

## Who to Notify

@brandenrodgers @camden11 @kemmerle


lib/fs.ts
```diff
@@ -69,6 +68,6 @@ export async function walk(dir: string): Promise<Array<string>> {
     .then(processFiles)
     .then(flattenAndRemoveSymlinks)
     .catch(err => {
-      throwError(err);
+      throw err;
```
joe-yeager, line 71: TODO: Maybe this should be a `FileSystemError`

brandenrodgers: Yeah that sound reasonable to me 👍 I think that's a safe change

lib/personalAccessKey.ts
```diff
@@ -48,21 +43,7 @@ export async function getAccessToken(
...
-      throwAuthErrorWithMessage(
-        `${i18nKey}.errors.invalidPersonalAccessKey`,
-        { errorMessage: error.response.data.message || '' },
-        error
-      );
-    } else {
-      throwError(e);
-    }
-  }
+  const response = await fetchAccessToken(personalAccessKey, env, accountId);
```
brandenrodgers, line 46: This is a great example of the value of this 🎉

lib/sandboxes.ts
```diff
@@ -27,26 +26,18 @@ export async function createSandbox(
   sandbox: Sandbox;
   personalAccessKey: string;
 }> {
```
brandenrodgers, line 28: We may need to run it by the sandboxes team, but we could probably just get rid of this file entirely, and the code in the CLI can call the request utils directly

models/HubSpotHttpError.ts
```diff
@@ -1,41 +1,239 @@
...
+      generatedContext.request = cause.config?.url;
+    } catch (e) {
+      logger.debug(e);
+    }
+
+    // Allow the provided context to override the generated context
+    this.context = { ...generatedContext, ...context };
+  }
+
+  private parseValidationErrors(
```
brandenrodgers, line 98: I think we might need to dig into this parse validation util. I noticed that we're logging out some of the important context, but not all of it. For example:

Here's what logs out when I get a "missing scopes" error response. I can get this by trying to run this command against my developer portal, which doesn't have the right content scopes to access CMS data.
image

It's trying to tell me the scopes I'm missing, but it's not including all of the information that I need. This is the shape of `responseData` when that happens:
image

`context`  here contains the scopes, like this: `["content", "content"]` (not sure why it's in there twice)

---

Also, from what I can tell, in the previous version of the error handling [we only called it in a few places ](https://github.com/search?q=repo%3AHubSpot%2Fhubspot-local-dev-lib%20throwApiUploadError&type=code) (throwApiUploadError is the only util that called isApiUploadValidationError,  which ran this validation parsing logic). I wonder if it will have an impact on some of the existing error messages?

models/HubSpotHttpError.ts
```diff
@@ -1,41 +1,239 @@
...
+
+    if (context.accountId) {
+      const action =
+        (method && HTTP_METHOD_VERBS[method]) || HTTP_METHOD_VERBS.get;
+
+      const preposition =
+        (method && HTTP_METHOD_PREPOSITIONS[method]) ||
+        HTTP_METHOD_PREPOSITIONS.get;
+
+      const requestName = context.request
```
brandenrodgers, line 149: Funny enough, I think you actually fixed this logic. I'm not sure which version I like better. Showing them the exact endpoint that we failed to hit seems like it might be more information than the end user actually needs. But maybe this is a case where the CLI should be intercepting this and logging something more user-friendly to the end user?

image

camden11: I think if we think about in in the context of the local-dev-lib end user vs the CLI end user it could make sense to do it this way

camden11 (commented):
This is looking great, really cleans things up nicely and should hopefully make the experience of using and contributing to local-dev-lib a lot better. I haven't tested yet, but I did my best to look over everything and didn't notice any problems. Are there any particular areas you need help with extra testing or that I should give a more focused look?

models/HubSpotHttpError.ts
joe-yeager: I agree with Camden that from the library perspetive it makes sense to give as informative of error messages as possible.  I think the problem actually came do to me trying to get clever and auto generate the context object from the `AxiosError`, which always gives us maximum detail. I think I found a good middle ground of having `message` in the `HubSpotHttpError` be a high level message that only has the `accountId` for the context.  And then having `internalMessage` (although maybe `detailedMessage` is a better name), be the message with the full context.  The `internalMessage` is what gets logged out when we call the `toString` method.

This gives us the flexibility to add the context when wanted to user facing errors, while making sure that we still capture the full error for debugging.

joe-yeager:
> Are there any particular areas you need help with extra testing or that I should give a more focused look?

Since it is such a wide spreading change, I think we really need to try to touch as many places as possible, so just widespread general testing.

lib/sandboxes.ts
joe-yeager: I wrote an issue and added it to our board to follow up on removing this file

models/HubSpotHttpError.ts
joe-yeager: I added a method to log the [formattedValidationErrors](https://github.com/HubSpot/hubspot-local-dev-lib/pull/159/files#diff-ae25cee36b980aa393de912585da6d585d226617b9595f4a65e6e3215c2a1925R93-R98) that the [CLI now calls if the error is a validation error.](https://github.com/HubSpot/hubspot-cli/pull/1109/files#diff-7fef93fbb06f6ac57ebe30f3ac8e8ab4fd9ececfbf84c740d5a255d1ff1c1cb2R25-R27)

I'll look into using the context object.  How did you trigger missing scopes error?  Just remove those scopes from the app?

camden11 (approved):
I still haven't had a chance to test this much, but assuming there will be plenty of time to text the `next` branch if its not going to be merged for a bit.

Looked everything over again and LGTM

models/HubSpotHttpError.ts
brandenrodgers: The reference to "app" here is misleading. In reality the error is referencing the personal access key. So you can generate a personal access key that doesn't have any scopes, and then try to run various CLI commands.

brandenrodgers (approved):
This lgtm! I did some testing locally and it all seems to be WAD. We'll have to hammer on it some more since it has a wide impact, but this is good to go into the `next` branch 👌

PR #161, opened 2024-07-18
Export getAxiosConfig from package

camden11:
## Description and Context
This exports `getAxiosConfig` from `local-dev-lib`. This will make it easier to make un-authed requests to HubSpot APIs, which is not currently supported by the `http` module.

More context: https://hubspot.slack.com/archives/C076QV345ME/p1721328218238189

## TODO
In the future, we should provide a way to make un-authed requests directly from the `http` module.

## Who to Notify

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "./api/*": "./api/*.js",
     "./errors/*": "./errors/*.js",
     "./http": "./http/index.js",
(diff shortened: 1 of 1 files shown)


kemmerle (approved):
:shipit:

joe-yeager (approved):

PR #162, opened 2024-07-23
Add a method to get available scopes on pak 

wung-hubspot:
## Description and Context
Adds a method that grabs the scopes on an access token for a given account.

### DIFF
--- a/lib/personalAccessKey.ts
+++ b/lib/personalAccessKey.ts
@@ -140,12 +140,11 @@ async function getNewAccessTokenByAccountId(
     throwErrorWithMessage(`${i18nKey}.errors.accountNotFound`, { accountId });
   }
   const { auth, personalAccessKey, env } = account;
-  const authTokenInfo = auth && auth.tokenInfo;
 
   const accessTokenResponse = await getNewAccessToken(
     accountId,
     personalAccessKey,
-    authTokenInfo && authTokenInfo.expiresAt,
+    auth?.tokenInfo?.expiresAt,
     env
   );
   return accessTokenResponse;
--- a/lib/personalAccessKey.ts
+++ b/lib/personalAccessKey.ts
@@ -184,7 +184,7 @@ export async function enabledFeaturesForPersonalAccessKey(
   return accessTokenResponse?.enabledFeatures;
 }
 
-export async function scopeGroupsForPersonalAccessKey(
+export async function scopesOnAccessToken(
   accountId: number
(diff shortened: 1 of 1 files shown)


lib/personalAccessKey.ts
```diff
@@ -132,6 +132,25 @@ async function getNewAccessToken(
...
 
+async function getNewAccessTokenByAccountId(
+  accountId: number
+): Promise<AccessToken> {
+  const account = getAccountConfig(accountId) as PersonalAccessKeyAccount;
+  if (!account) {
+    throwErrorWithMessage(`${i18nKey}.errors.accountNotFound`, { accountId });
+  }
+  const { auth, personalAccessKey, env } = account;
+  const authTokenInfo = auth && auth.tokenInfo;
```
kemmerle, line 143: I think it'd be good to wrap this in a try...catch block, in case `getNewAccessToken` fails. Also, what happens if `auth.tokenInfo` is falsey? Then `authTokenInfo` will be `null`--will `getNewAccessToken` accept `null`?

wung-hubspot: simplified it to a bunch of coalesces, it's fine if expiresAt is not provided. 

I'm not sure the benefits of grouping this in a try catch is, could you elaborate?

kemmerle: @wung-hubspot, just saw this. I looked at `getNewAccessToken`, where there is a try catch block, so ignore this comment. Going to approve this now.

kemmerle (approved):
LGTM

lib/personalAccessKey.ts
```diff
@@ -161,22 +179,16 @@ export async function accessTokenForPersonalAccessKey(
...
-  const { auth, personalAccessKey, env } = account;
-  const authTokenInfo = auth && auth.tokenInfo;
-
-  const accessTokenResponse = await getNewAccessToken(
-    accountId,
-    personalAccessKey,
-    authTokenInfo && authTokenInfo.expiresAt,
-    env
-  );
+  const accessTokenResponse = await getNewAccessTokenByAccountId(accountId);
```
brandenrodgers, line 182: Rather than exposing one-off utils to access fields on the access token response, I think we might be better off just exposing a util to get the access token for a given account id. Then we won't have to come in here and expose new utils every time we want to access a new field on the access token response. 

We'll probably have to keep `enabledFeaturesForPersonalAccessKey` around though, just to prevent this form becoming a breaking change.

PR #163, opened 2024-07-23
Allow asking for full response from http methods

bmatto:
## Description and Context

We need to have the full response from Axios for a use-case internally where we need to pull headers off of the incoming request.

### DIFF
--- a/http/index.ts
+++ b/http/index.ts
@@ -121,49 +121,81 @@ function addQueryParams(
 
 async function getRequest<T>(
   accountId: number,
-  options: HttpOptions
-): Promise<T> {
+  options: HttpOptions,
+  withFullResponse = false
+): Promise<T | AxiosResponse> {
   const { params, ...rest } = options;
   const axiosConfig = addQueryParams(rest, params);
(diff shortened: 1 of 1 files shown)


camden11 (commented):
@bmatto Using overloads is more verbose but would let us keep the types accurate

```typescript
async function getRequest<T>(
  accountId: number,
  options: HttpOptions,
  withFullResponse: false
): Promise<T>;
async function getRequest<T>(
  accountId: number,
  options: HttpOptions,
  withFullResponse: true
): Promise<AxiosResponse<T>>;
async function getRequest<T>(
  accountId: number,
  options: HttpOptions,
  withFullResponse?: boolean
) {
  const { params, ...rest } = options;
  const axiosConfig = addQueryParams(rest, params);
  const configWithAuth = await withAuth(accountId, axiosConfig);

  const response = await axios<T>(configWithAuth);

  if (withFullResponse) {
    return response;
  }

  return response.data;
}
```

camden11 (approved):
LGTM - thanks for making that update

PR #164, opened 2024-07-24
Add API function to get public app install counts without delay

camden11:
## Description and Context
This adds support for hitting the new `apps-dev/external/public/v3/{appId}/install-counts-without-test-portals` endpoint, which returns public app production installs without the 30 minute delay the endpoint we were previously using had. This will allow the CLI to give users more up to date info about whether or not their public apps have active installs

## Who to Notify

### DIFF
--- a/api/appsDev.ts
+++ b/api/appsDev.ts
@@ -1,6 +1,7 @@
 import http from '../http';
 import {
   PublicApp,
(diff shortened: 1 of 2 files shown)


joe-yeager (approved):
LGTM

PR #165, opened 2024-07-24
fix!: Remove default exports and add linting rule for default exports.

joe-yeager:
## Description and Context
_**Note: This will be a breaking change, so this PR is targeting the next branch and will be in the next major**_

- Remove default exports and add a linting rule to prevent usage.

## Who to Notify

@camden11 @brandenrodgers @kemmerle

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,11 +12,12 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2022,
   },
-  plugins: ['@typescript-eslint'],
+  plugins: ['@typescript-eslint', 'import'],
   rules: {
     'no-console': 'off',
     'no-return-await': 'error',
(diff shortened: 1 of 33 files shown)


camden11 (approved):
Nice! LGTM. Will just have to remember each of these changed exports so we can be sure to replace them properly in the CLI

joe-yeager:
> Nice! LGTM. Will just have to remember each of these changed exports so we can be sure to replace them properly in the CLI

Good callout.  I should have linked the PRs, but I have a [PR to update the CLI](https://github.com/HubSpot/hubspot-cli/pull/1117/files).  I was only able to find one usage

kemmerle (approved):
Reviewed these changes, as well, and tried out commands locally with these changes. Everything's working for me.

PR #167, opened 2024-07-26
Expose axios response

camden11:
## Description and Context
This updates the `http` module to return full Axios response objects, which will make debugging easier for libraries that consume locla-dev-lib. All API modules will also now return the full response object (they've been updated to use the `AxiosPromise` return type that `axios` includes. Their previous return values will now be within the `data` field on the response object.

A Few other things...
* Uses this as an opportunity to clean up the API modules a bit. Added some return types and removed a lot of unneeded `async`s.
* Added a testing util to create Axios responses in `__utils__/mockAxiosResponse`. This lets tests easily create type accurate axios responses and use them for mocked return values. There _might_ be a better way to do this, but figured this would get the job done for now.

## TODO
Need to be very careful to make update all uses of API modules in the CLI to make sure nothing breaks. May also be helpful to PR other HubSpot repos that use local-dev-lib, if needed

## Who to Notify

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosResponse } from 'axios';
+import axios, { AxiosPromise } from 'axios';
 import { getDefaultUserAgentHeader } from '../http/getAxiosConfig';
 import { GithubReleaseData, GithubRepoFile } from '../types/Github';
 
@@ -22,7 +22,7 @@ const GITHUB_AUTH_HEADERS = {
 export async function fetchRepoReleaseData(
   repoPath: RepoPath,
   tag = ''
-): Promise<AxiosResponse<GithubReleaseData>> {
+): AxiosPromise<GithubReleaseData> {
   const URL = `${GITHUB_REPOS_API}/${repoPath}/releases`;
 
   return axios.get<GithubReleaseData>(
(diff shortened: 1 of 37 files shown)


api/projects.ts
```diff
@@ -275,36 +276,40 @@ export async function cancelStagedBuild(
   });
 }
 
-export async function fetchBuildWarnLogs(
+type WarnLogsResponse = {
+  logs: Array<ProjectLog>;
+};
```
joe-yeager, line 281: Do we want to move this to a types file?

joe-yeager (approved):
LGTM!  Just one minor TS question

api/projects.ts
camden11: Hmm good question. For smaller types that are only used in one file, I generally just keep them in that file. Though thinking from a library perspective it would probably make sense to export all those types.

This is done in a lot of places though so would need a separate PR

api/github.ts
```diff
@@ -35,9 +35,7 @@ export async function fetchRepoReleaseData(
 
 // Returns the entire repo content as a zip, using the zipball_url from fetchRepoReleaseData()
 // https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-zip
-export async function fetchRepoAsZip(
-  zipUrl: string
-): Promise<AxiosResponse<Buffer>> {
+export async function fetchRepoAsZip(zipUrl: string): AxiosPromise<Buffer> {
```
brandenrodgers, line 38: Can we remove the async from these ones too?

lib/developerTestAccounts.ts
```diff
@@ -15,8 +15,8 @@ export async function createDeveloperTestAccount(
   accountName: string
 ): Promise<DeveloperTestAccount> {
   try {
-    const resp = await _createDeveloperTestAccount(accountId, accountName);
-    return resp;
+    const { data } = await _createDeveloperTestAccount(accountId, accountName);
+    return data;
```
brandenrodgers, line 19: I mentioned this to @joe-yeager, but these utils are such thin wrappers now that I think we should just delete the entire file. It's not really adding much value at all. We could probably delete lib/sandboxes too, since it's also not doing any additional logic in the api call wrappers.

We can do this in another PR though. It's possible Joe already did it in his error handling PR

brandenrodgers (approved):
Left some small comments, but otherwise lgtm!

lib/developerTestAccounts.ts
joe-yeager: I haven't done it yet, but I'll add an issue so we don't lose track of it and do it in a follow up PR.

PR #169, opened 2024-07-30
Remove custom project 403 error handlers

brandenrodgers:
## Description and Context

We don't need this custom error handling anymore. In some cases (i'll explain below) the errors that we're throwing are actually incorrect. [We had added these custom overrides a while ago](https://github.com/HubSpot/cli-lib/pull/4) because the errors we were getting from the BE were not containing enough useful information. The gating and scoping errors returned by the BE are much more informative now, so we don't need these overrides anymore. Now we will use our default error handling logic provided by the api utils in this library.

We may want to have our designers take a look at the new errors just to make sure we're happy with them.

#### Before
image

**NOTE:** This is actually incorrect. This error is referencing the private app public beta.

#### After
image

---
#### Before
image

#### After
image

**NOTE:** There is no change in behavior here. This is because the error returned by the BE does not have a `category` of `MISSING_SCOPES` or `GATED`, which our overrides rely on.

---
#### Before
image

#### After
image

---
#### Before
image

**NOTE:** This is actually incorrect. This error is referencing the private app public beta.

#### After
image

---
#### Before
image

#### After
image

## TODO

- [ ] Verify error copy with designers

## Who to Notify

cc/ @m-roll


sandracoleto:
### 1.  Uploading a public app to a non-developer account

**Suggestion**
x Failed to upload XXX project files to XXX
[ERROR] Your project file has a [public app] component that's only supported by developer accounts. Your current default account is a production account.  Run 'hs accounts use' to change your default account. Learn more about public apps in developer accounts (link to documentation).

**If there is not link to docs.** 
x Failed to upload XXX project files to XXX
[ERROR] Your project file has a [public app] component that's only supported by developer accounts. Your current default account is a production account.  Run 'hs accounts use' to change your default account. To learn more about public apps in developer accounts contact support. 

### 2. Uploading a private app to a developer account
**Suggestion**
x Failed to upload XXX project files to XXX
[ERROR] Your project file has a [private app] component that's only supported by production accounts, standard sandboxes, and development sandboxes. Your current default account is a developer account.  Run 'hs accounts use' to change your default account. Learn more about private apps (link to doc).
https://developers.hubspot.com/docs/platform/create-private-apps-with-projects

### 3. Uploading a private app project to a personal account when not ungated for the crm dev beta
**Suggestion**
x Failed to upload XXX project files to XXX
[ERROR] The current default account, XXX, doesn’t have access to CRM Development Tools beta. Learn more about and join the beta here <link to the public beta enrollment page>.

### 4. Uploading a public app project to a developer account when not ungated for the public app beta
**Suggestion**
x Failed to upload XXX project files to XXX
[ERROR] The current default account, XXX, doesn’t have access to the Developer Projects features. To learn more about Developer Projects in developer accounts contact support.

### 5. Uploading a private app project to a personal account when the PAK has not been granted the proper scopes
**Question**
Can we identify the missing scopes here?  and say something like 
"Unable to use private app user token for appId~{{ appId }} since **App function** scope is missing on accountId~{{ accountId }}. To add App function scope to that account, deactivate the existing Personal Access Key, and generate a new one that includes the App functions scope. Then run hs auth and reauthenticate accountId~{{ accountId }}.  ?

brandenrodgers:
## Status update on 403 errors

### 1.  Uploading a public app to a non-developer account
image

**This should be:**
`
[ERROR] Your project file has a [public app] component that's only supported by developer accounts. Your current default account is a production account. Run 'hs accounts use' to change your default account. Learn more about public apps in developer accounts (link to documentation).
`

### 2. Uploading a private app to a developer account
image

**This should be:**
`
[ERROR] Your project file has a [private app] component that's only supported by production accounts, standard sandboxes, and development sandboxes. Your current default account is a developer account. Run 'hs accounts use' to change your default account. Learn more about private apps (link to doc).
`

### 3. Uploading a private app project to a personal account when not ungated for the crm dev beta
image

**This should be:**
`

`

### 4. Uploading a public app project to a developer account when not ungated for the public app beta
image

**This should be:**
`

`
### 5. Uploading a private app project to a personal account when the PAK has not been granted the proper scopes
image

camden11 (approved):

brandenrodgers:
### Status update 2 on errors

### 1.  ❌ Uploading a public app to a non-developer account
image

**NOTE:** This error is misleading. It makes no mention of account types, and it tells the user to reach out to support. The real fix here is for them to switch to a developer account and re-upload.

### 2. ❌ Uploading a private app to a developer account
image

**NOTE:** The BE error is getting overridden here because we're overriding anything with a category of `MISSING_SCOPES`. The BE that we're getting now seems like a regression from the behavior in my previous status update.

### 3. ⚠️ Uploading a private app project to a personal account when not ungated for the crm dev beta

**NOTE:** For some reason I'm not able to trigger this one anymore. I gated my portal for `crm-development-beta`, but Projects UI is still loading and project uploads are still succeeding.

### 4. ⚠️ Uploading a public app project to a developer account when not ungated for the public app beta
image

**NOTE:** This isn't awful, but it would be better if we could directly link to the beta that the user needs to opt into

### 5. ✅ Uploading a private app project to a personal account when the PAK has not been granted the proper scopes
image

PR #171, opened 2024-08-02
Unauthed http option

camden11:
## Description and Context
This adds new methods for hitting hubspot endpoints that don't require auth. See thread for more details on this ask:
https://hubspot.slack.com/archives/C076QV345ME/p1721328218238189

Decided it would be easier to put this into the `next` branch due to the changes to `http`, even though it doesn't actually have any breaking changes.

## TODO
I feel like there might be a better way to name these functions? I couldn't think of anything that I really liked though.

## Who to Notify

### DIFF
--- a/http/index.ts
+++ b/http/index.ts
@@ -8,7 +8,7 @@ import { USER_AGENTS, getAxiosConfig } from './getAxiosConfig';
 import { accessTokenForPersonalAccessKey } from '../lib/personalAccessKey';
 import { getOauthManager } from '../lib/oauth';
 import { FlatAccountFields } from '../types/Accounts';
-import { AxiosConfigOptions, HttpOptions, QueryParams } from '../types/Http';
+import { HttpOptions, QueryParams } from '../types/Http';
 import { logger } from '../lib/logger';
 import { i18n } from '../utils/lang';
 import { HubSpotHttpError } from '../models/HubSpotHttpError';
@@ -79,7 +79,7 @@ function withPortalId(
 
 async function withAuth(
   accountId: number,
-  options: AxiosConfigOptions
+  options: HttpOptions
 ): Promise<AxiosRequestConfig> {
   const accountConfig = getAccountConfig(accountId);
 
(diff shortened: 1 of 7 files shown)


http/index.ts
```diff
@@ -130,43 +130,70 @@ async function getRequest<T>(
...
-  const configWithAuth = await withAuth(accountId, options);
+  const requestConfig = await withAuth(accountId, options);
+  return axios({ ...requestConfig, method: 'post' });
+}
 
-  return axios({ ...configWithAuth, method: 'post' });
+async function unauthedPostRequest<T>(options: HttpOptions): AxiosPromise<T> {
+  const requestConfig = await getAxiosConfig(options);
+  return axios({ ...requestConfig, method: 'post' });
 }
```
joe-yeager, line 158: Is the reason for splitting this into a new method that the callers might now have the `accountId` in context?  If so, what would you think about updating the original methods to take a single object, and make `accountId` optional.  Then if the `accountId` is provided, we generate the auth and make the unauthed request otherwise.

camden11: Yeah, you don't need an `accountId` to make an unauthed request, so I figured we should assume in those cases that it won't exist. Was trying to approach this in a way that wouldn't make any API changes but I guess since we have the major release going out anyway there's no downside to doing that now. Hmm.

I could go either way. Would definitely clean things up a bit, but I feel like you could also make the case that the vast majority of hubspot endpoints do require auth (and in our case, every single one the CLI interacts with), and moving `accountId` into `options` could obscure the fact that it's necessary for most requests

camden11:
Something @brandenrodgers and I just talked about in our 1:1 is moving the unauthed methods to their own file but having them use the standard names.

so something like `import { get } from '@hubspot/local-dev-lib/http/unauthed'`

kemmerle:
@camden11, I agree with Branden that moving these methods into their own file and using standard names is more streamlined.

http/index.ts
joe-yeager: I just saw your comment about Branden's suggestion and I like that as well and it addresses the core of my concern, so I'll flag this resolved.

brandenrodgers (approved):
👌 nice I like this unauthed pattern

PR #172, opened 2024-08-05
Refactor(config) : Sister PR in hubspot-cli

kemmerle:
## Description and Context
We are updating our HubSpot CLI configuration file, so that it will be created the root directory in a hidden folder: `~/.hubspot-cli/config.yml`. This will encourage users to take advantage of the `init`, `auth`, and `accounts` commands to update their config and help prevent config-related bugs. 

We will also be supporting the old configuration, which means that power users will still be able to use nested configs for different projects.

Sister PR: https://github.com/HubSpot/hubspot-cli/pull/1123

## QUESTIONS:

For Sandra, how would we like to alert customers about the new config? Currently, I have a prompt that asks them whether they'd like to use the new vs. old, but I'm not sure that's the best solution.

## Screenshots

`hs init` (with new config):

Screenshot 2024-08-05 at 11 34 19 AM

`hs init` (Erroring out because config exists):

This isn't the greatest, because the prompt for new vs old config still appears.

Screenshot 2024-08-05 at 11 39 57 AM

`hs auth` (Adding another account):

Screenshot 2024-08-05 at 11 34 59 AM

## TODO

- [x] Need to fix minor bug where choose default account prompt still appears when re-authing a default account.
- [x] Need to check for both old and new config paths to prevent users from using both config types. 
- [x] Add tests

## Who to Notify


config/index.ts
```diff
@@ -81,11 +81,11 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
...
 }
 
-export function getConfigPath(path?: string): string | null {
-  if (CLIConfiguration.isActive()) {
+export function getConfigPath(useNewConfig = false): string | null {
+  if (useNewConfig || CLIConfiguration.isActive()) {
     return getConfigFilePath();
   }
-  return config_DEPRECATED.getConfigPath(path);
+  return config_DEPRECATED.getConfigPath();
```
camden11, line 88: You can probably update the deprecated config and remove this arg now that its not being used anywhere

constants/config.ts
```diff
@@ -2,7 +2,8 @@ import { i18n } from '../utils/lang';
 
 export const DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME = 'hubspot.config.yml';
 
-export const HUBSPOT_CONFIGURATION_FOLDER = '.hubspot';
+// Changed for easier debugging.
+export const HUBSPOT_CONFIGURATION_FOLDER = 'hubspot';
```
camden11, line 6: Is this just a temporary change?

config/index.ts
```diff
@@ -81,11 +81,11 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
   }
 }
 
-export function getConfigPath(path?: string): string | null {
-  if (CLIConfiguration.isActive()) {
+export function getConfigPath(useNewConfig = false): string | null {
```
camden11, line 84: This is also _technically_ a breaking change even though it doesn't look like we actually pass the string in anywhere. If possible might be best to find a way to do this without breaking changes. Maybe adding a new method?

constants/config.ts
kemmerle: Temporary change, no worries.

config/CLIConfiguration.ts
```diff
@@ -346,10 +346,13 @@ class CLIConfiguration {
...
+    // Check whether the account is already listed in the config.yml file.
     const currentAccountConfig = this.getAccount(accountId);
 
+    // For accounts that are already in the config.yml file, sets the auth property.
     let auth: OAuthAccount_NEW['auth'] =
       (currentAccountConfig && currentAccountConfig.auth) || {};
+    // For accounts not already in the config.yml file, sets the auth property.
     if (clientId || clientSecret || scopes || tokenInfo) {
       auth = {
         ...(currentAccountConfig ? currentAccountConfig.auth : {}),
```
brandenrodgers, line 358: Is this line equivalent to `...auth` because of the logic right above it?

config/CLIConfiguration.ts
```diff
@@ -405,25 +408,23 @@ class CLIConfiguration {
...
+    if (!Object.hasOwn(this.config, 'accounts')) this.config.accounts = [];
     if (currentAccountConfig) {
       logger.debug(
         i18n(`${i18nKey}.updateAccount.updating`, {
           accountId,
         })
       );
       const index = this.getConfigAccountIndex(accountId);
-      this.config.accounts[index] = completedAccountConfig;
+      if (index < 0) this.config.accounts.push(completedAccountConfig);
```
brandenrodgers, line 419: If this util is called `updateAccount()`, what's the scenario where we'd call it without the account already existing in the config?

I poked around in this file and it looks like there's no `addAccount()` util. Maybe we should name this one `addOrUpdateAccount()` for clarity? This is just a nit.

config/CLIConfiguration.ts
```diff
@@ -405,25 +408,23 @@ class CLIConfiguration {
     safelyApplyUpdates('parentAccountId', parentAccountId);
 
     const completedAccountConfig = nextAccountConfig as FlatAccountFields_NEW;
-
+    if (!Object.hasOwn(this.config, 'accounts')) this.config.accounts = [];
```
brandenrodgers, line 411: Another slight nit (and maybe just personal preference), but mixing the inline ifs with the bracketed ones makes this a little harder to read. I'd be fine with using brackets here to make it easier to follow, even if it adds a few lines.

lib/personalAccessKey.ts
```diff
@@ -245,11 +246,13 @@ export async function updateConfigWithAccessToken(
     parentAccountId,
     env: accountEnv,
   });
-  writeConfig();
+  if (!CLIConfiguration.isActive()) {
```
brandenrodgers, line 249: Why do we only want to write to the config here when the new version isn't active?

config/CLIConfiguration.ts
kemmerle: As I understand it, yes.

lib/personalAccessKey.ts
kemmerle: In the new version, we write the config directly in `CLIConfiguration`.

config/index.ts
kemmerle: There is still one method in `config_DEPRECATED` that uses this function with the path argument: `loadConfigFromFile`. Going to keep as is for now but let me know if you have any ideas/thoughts.

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/0559c584504dc458721cef1dced436eaf0e88862

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/0559c584504dc458721cef1dced436eaf0e88862

constants/config.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/0559c584504dc458721cef1dced436eaf0e88862

config/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/b8c4b88463b52a4faf4448223fe68e58064d52af

kemmerle:
@joe-yeager, I could use your help with the tests that I added for the `getRootOrDeprecatedConfigPath` and `bothConfigFilesExist` functions. I would like to test out the `config_DEPRECATED.getConfigPath` method as part of these tests, but importing and mocking the file in the tests results in them breaking, I believe because there is already a `getConfigPath` function imported from the `index` file. Even importing the deprecated function under an alias does not work. 

I am going to merge some commented out code--would you take a closer look async? Or I'm happy to hop in a huddle today or tomorrow.

config/__tests__/config.ts
```diff
@@ -32,6 +34,8 @@ import {
   PersonalAccessKeyAccount,
   PersonalAccessKeyAccount_DEPRECATED,
 } from '../../types/Accounts';
+import * as configFile from '../configFile';
+//import * as config_DEPRECATED from '../config_DEPRECATED';
```
kemmerle, line 38: @joe-yeager, see my note. I've added the `config_DEPRECATED` tests for your review. Any help is appreciated!

config/index.ts
```diff
@@ -81,13 +81,38 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
...
+    return true;
+  }
+
+  if (!useRootConfig && configFileExists()) {
+    return true;
+  }
+
+  return false;
+}
+
```
camden11, line 115: I think there may be a flaw in the logic here. Or at least, it's not actually checking if both config files exist, in which case, it might make sense to rename it.

The first check for both config files existing effectively isn't doing anything (other than short circuiting, but I don't think thats really worth it at the expense of the logic being harder to follow), since if both exist, one of the following conditions will always be true.

If the logic here is working as intended, I _think_ this could be simplified to
```
return useRootConfig ? Boolean(config_DEPRECATED.getConfigPath()) : configFileExists()
```

camden11: Based on how this is used in `hs auth`, it looks like it'll return true whenever `configFileExists()` is true

config/index.ts
```diff
@@ -81,13 +81,38 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
   }
 }
 
+export function getRootOrDeprecatedConfigPath(
```
brandenrodgers, line 84: Not a blocker, but is there a better term for the new config than "root"? I realize that I'm probably the one who initially started using "root", but it actually might be slightly misleading. The location where we're actually placing the config file is in the user's home directory (i.e. `~/`). Just throwing out some other options:
- Central config
- Single config
- hidden config

lib/personalAccessKey.ts
brandenrodgers: We may want to add a quick comment in here to explain that, just so that someone who stumbles onto this will understand

config/__tests__/config.ts
kemmerle: Resolving for now--Joe reviewed and left me a comment on Slack.

config/index.ts
kemmerle: Thank you, Camden, this was really helpful and helped me get my head on straight.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/c14b1aa085ea0995951a0a73da3599756b91415b

config/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/c14b1aa085ea0995951a0a73da3599756b91415b

constants/config.ts
```diff
@@ -2,7 +2,8 @@ import { i18n } from '../utils/lang';
 
 export const DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME = 'hubspot.config.yml';
 
-export const HUBSPOT_CONFIGURATION_FOLDER = '.hubspot';
+// Temporary change for easier debugging.
+export const HUBSPOT_CONFIGURATION_FOLDER = 'hubspot';
```
camden11, line 6: Should this be switched back before merging?

camden11 (approved):
One question + looks like you have a conflict, but otherwise LGTM

joe-yeager (approved):

config/index.ts
```diff
@@ -47,9 +47,9 @@ export function loadConfigFromEnvironment(): boolean {
 
 export function createEmptyConfigFile(
```
brandenrodgers, line 48: Should this util respect the `path` option? If the user specifies a custom path, they're attempting to tell us that they want to create/read a config that lives somewhere different from the default location. I think I had unintentionally broken this in my initial pass at implementing this new config behavior.

config/index.ts
```diff
@@ -81,13 +81,32 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
   }
 }
 
+export function getHiddenOrDeprecatedConfigPath(
```
brandenrodgers, line 84: Do we need this? Is it any different from what `getConfigPath()` is doing? What if we defined getConfigPath as `getConfigPath(path?: string, useHiddenConfig?: boolean)`. Every util in this file works for both the deprecated and hidden config, so including the `getHiddenOrDeprecated` part in the name isn't necessary.

config/index.ts
```diff
@@ -81,13 +81,32 @@ export function writeConfig(options: WriteConfigOptions = {}): void {
...
+}
+
 export function getConfigPath(path?: string): string | null {
   if (CLIConfiguration.isActive()) {
     return getConfigFilePath();
   }
   return config_DEPRECATED.getConfigPath(path);
 }
 
+export function otherConfigFileExists(useHiddenConfig?: boolean): boolean {
```
brandenrodgers, line 100: Could both of these utils be simplified into a single export of `configFileExists`. For example:
```js
export function configFileExists(usehiddenConfig?: boolean) {
    return useHiddenConfig
    ? configFile.configFileExists()
    : Boolean(config_DEPRECATED.getConfigPath())
}

// Then used in CLI like...

// Checking if the other exists:
const doesOtherConfigFileExist = configFileExists(!useHiddenConfig)

// Checking if both exist
const bothExist = configFileExists(true) && configFileExists(false)
```

Then the logic in our export doesn't need to be inverted (i.e. "other"), and also we'd only need a single export

config/index.ts
kemmerle: In my original pass, I deleted the path argument from `getConfigPath` function in this file, and @camden11 pointed out that that would be a breaking change. I created a separate function to avoid that, but this is the simpler solution 🤦‍♀️

config/index.ts
kemmerle: I've fixed this behavior for the hidden config. 

We only call this function in the `init` command, and if the `useHiddenConfig` flag is used, we run `CLIConfiguration.write({ accounts: [] })`, which doesn't use the `path` option at all. Only the deprecated config respects it, which we want.

config/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/b40fdfb4fe2448ef929b7d50b11bcb1fca8d55ab

config/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/e50aca5114cdd58c397440f70c645a25cdc59857

constants/config.ts
brandenrodgers: Just a reminder to switch this back to a dot file. I also think we may benefit from increasing the specificity here to something like `.hubspot-cli`. Otherwise it'll conflict with existing files on internal user's machines

joe-yeager: @kemmerle Does this need to be done as a part of this PR?

config/CLIConfiguration.ts
```diff
@@ -257,13 +257,29 @@ class _CLIConfiguration {
...
+        this.config.accounts.find(account => account.accountId === accountId) ||
+        null
+      );
+    }
+    return null;
+  }
+
+  getConfigAccounts(): Array<CLIAccount_NEW> | null {
+    if (this.config) {
+      return this.config.accounts || null;
```
camden11, line 270: mega nit: for this and `getConfigForAccount` above you can use the `?.` operator to avoid having to do the `if` statement. There are maybe a few other places where this is applicable

Instead of
```
if (this.config) {
      return this.config.accounts || null;
}
return null
```
it would be
```
return this.config?.accounts || null
```

config/CLIConfiguration.ts
```diff
@@ -257,13 +257,29 @@ class _CLIConfiguration {
...
     }
     return null;
   }
 
   isAccountInConfig(nameOrId: string | number): boolean {
+    if (typeof nameOrId === 'string') {
+      return (
+        !!this.config &&
+        this.config.accounts &&
```
camden11, line 279: Here's another place `?.` could come in handy

camden11 (approved):
Just one minor style suggestion (feel free to take it or leave it) otherwise everything LGTM

PR #173, opened 2024-08-05
Update some endpoints to point towards new project's service routes

Triippz:
## Description and Context

We have moved over a number of REST endpoints from the build service to the projects service in an attempt to separate out functionality. This PR updates a handful of those endpoints which have changed. Nothing about the endpoints has changed outside of the new routing, so request/responses should be exactly the same.

## TODO

We will be maintaining the old REST endpoints on the build service for the time being but we will need to coordinate when we can deprecate those without impacting users with older CLI versions. We might be able to utilize what @m-roll made a few weeks ago using the cli version headers to block requests.
## Who to Notify

cc @brandenrodgers @ypzheng

### DIFF
--- a/api/projects.ts
+++ b/api/projects.ts
@@ -30,7 +30,7 @@ export async function fetchProjects(
   accountId: number
 ): Promise<FetchProjectResponse> {
   return http.get(accountId, {
-    url: PROJECTS_API_PATH,
+    url: DEVELOPER_PROJECTS_API_PATH,
   });
 }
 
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):

ypzheng:
Thanks. Sorry, before merging, we should also update the "fetchProjectSettings". I forgot to list it in the original issue because it was very dated.

ypzheng (changes requested):
Putting up a blocker just to make sure we don't miss it before merging since it was not listed on the original issue.

PR #176, opened 2024-08-07
Corrected - New Project Service Paths

Triippz:
## Description and Context

Same as https://github.com/HubSpot/hubspot-local-dev-lib/pull/173 but with corrected path.
## Who to Notify

@brandenrodgers -> please test
cc @ypzheng


camden11 (approved):
Tested all of these and works as expected.

PR #178, opened 2024-08-13
chore: deprecate `lib/developerTestAccounts.ts`

joe-yeager:
## Description and Context

- Deletes `lib/developerTestAccounts.ts` since it is a very thin wrapper over `api/developerTestAccounts` and we can just directly access those methods.

Companion PR: https://github.com/HubSpot/hubspot-cli/pull/1126

## Who to Notify

@kemmerle @camden11 @brandenrodgers @adamawang


adamawang (approved):
👋

camden11 (approved):

PR #179, opened 2024-08-14
chore: Add .test extension prefix

joe-yeager:
## Description and Context

Prefix the extension with `.test` so that the repo is easier to navigate

## Who to Notify

@kemmerle @brandenrodgers @camden11


camden11 (approved):
beautiful

PR #180, opened 2024-08-15
chore: Remove redundant file

joe-yeager:
## Description and Context

I had the commit to delete `lib/sandboxes.ts` staged locally but didn't push it before merging the previous PR.  This is no longer used on the CLI

## Who to Notify

@camden11 @kemmerle @brandenrodgers @adamawang


adamawang (approved):
🥾

PR #182, opened 2024-08-15
Fix fetch for themes with JSR modules

jsines:
## Description and Context
Report in cms dev tooling channel from infra that elevate theme is silently erroring on `tsx, ts, jsx` files when trying to fetch the theme.

This just adds those to the allowlist when checking extensions (don't want to add them to the default allowlist since that runs for `hs watch` as well). Error message here already exists just doesn't get logged in this circumstance

### DIFF
--- a/lib/fileMapper.ts
+++ b/lib/fileMapper.ts
@@ -197,7 +197,7 @@ async function fetchAndWriteFileStream(
     logger.log(i18n(`${i18nKey}.skippedExisting`, { filepath }));
     return;
   }
-  if (!isAllowedExtension(srcPath, ['tsx', 'jsx', 'ts'])) {
+  if (!isAllowedExtension(srcPath, ['tsx', 'ts', 'jsx'])) {
     throwErrorWithMessage(`${i18nKey}.errors.invalidFileType`, { srcPath });
   }
   let node: FileMapperNode;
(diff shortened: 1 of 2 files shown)


lib/fileMapper.ts
```diff
@@ -197,7 +197,7 @@ async function fetchAndWriteFileStream(
     logger.log(i18n(`${i18nKey}.skippedExisting`, { filepath }));
     return;
   }
-  if (!isAllowedExtension(srcPath)) {
+  if (!isAllowedExtension(srcPath, ['tsx', 'ts', 'jsx'])) {
```
TanyaScales, line 200: This feeds back to this set of constants -- https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/constants/extensions.ts#L1 - I see the list is being used for the watch/uploadFile area as well, so I can see the short fix here since the file types aren't 100% supported around the board. Perhaps we can make another set there with the CMS React extensions to that we can continue to use them around the lib as any further work progresses -- then we don't have to peel this out down the line or remember where this is set.

jsines: Threw it into a const Set and pulled from there instead 👍

TanyaScales (approved):

PR #183, opened 2024-08-19
Jy/next branch fixes

joe-yeager:


models/HubSpotHttpError.ts
```diff
@@ -14,6 +14,7 @@ export class HubSpotHttpError<T = any> extends Error {
   public headers?: { [key: string]: unknown };
   public method: string | undefined;
   public context: HubSpotHttpErrorContext | undefined;
+  public derivedContext: HubSpotHttpErrorContext | undefined;
```
joe-yeager, line 17: Splitting the context derived into it's own field so when the provided context is updated we don't slip details into the log the user didn't intend.

PR #185, opened 2024-08-20
fix: Logging fixes

joe-yeager:
## Description and Context

Fixes the error where we weren't logging out the required scopes.
Fixes errors with the payloads getting logged

### Before
Screenshot 2024-08-20 at 2 35 45 PM

### After
Screenshot 2024-08-20 at 2 33 24 PM

## Who to Notify

@kemmerle @brandenrodgers @camden11

### DIFF
--- a/models/HubSpotHttpError.ts
+++ b/models/HubSpotHttpError.ts
@@ -140,7 +140,14 @@ export class HubSpotHttpError<T = any> extends Error {
     if (errors) {
       const specificErrors = errors.map(error => {
         let errorMessage = error.message;
-        if (error.errorTokens && error.errorTokens.line) {
+
+        if (error.context?.requiredScopes) {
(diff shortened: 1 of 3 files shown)


models/HubSpotHttpError.ts
```diff
@@ -116,7 +121,7 @@ export class HubSpotHttpError<T = any> extends Error {
     generatedContext.request = cause.config?.url;
 
     // Allow the provided context to override the generated context
-    this.context = { ...generatedContext, ...this.context };
+    this.derivedContext = { ...this.derivedContext, ...generatedContext };
```
joe-yeager, line 124: Splitting this into `derivedContext` to prevent collisions with the user provided context and it doesn't unintentionally slip unwanted details into the `message`.

camden11 (approved):

PR #186, opened 2024-08-23
Removing unused task status fetch

adamawang:
## Description and Context

The fetchTaskStatus util was used to populate the progress bar we showed when a sync was active. We've since removed the progress bar from the UI, and are now deprecating the sync command entirely so these can safely be removed. 

## Screenshots

n/a


camden11:
@adamawang This looks good, but since it's removing functionality from `local-dev-lib`, it'll have to go out as part of a new major version. Luckily, we have one of those planned right now. Everything that's going to go out as part of the next major version is on the `next` branch right now, so if you switch the base of this PR form `main` to `next`, it'll be good to go

camden11 (approved):

PR #188, opened 2024-08-28
Fix: Add function to interpret tilde in file paths

kemmerle:
## Description and Context
In the create project prompt, when a user is prompted to enter a project location, we are not interpreting the `~` correctly. I've added a util function that allows us to untildify the tilde. 

NOTE: For some reason, I received Prettier warnings about some files that @joe-yeager recently edited iirc. Changed them, let me know if I should revert. 

## Screenshots

Before (Note the weird location: `/Users/akemmerle/src/~/src/test-proj828`):

Screenshot 2024-08-28 at 4 29 58 PM

After:

Screenshot 2024-08-28 at 4 30 09 PM

## TODO

- [ ] Address feedback
- [ ] Merge local-dev-lib PR and update version in CLI. 

## Who to Notify

### DIFF
--- a/api/projects.ts
+++ b/api/projects.ts
@@ -76,7 +76,9 @@ export async function fetchProject(
   projectName: string
 ): Promise<Project> {
   return http.get(accountId, {
-    url: `${DEVELOPER_PROJECTS_API_PATH}/by-name/${encodeURIComponent(projectName)}`,
+    url: `${DEVELOPER_PROJECTS_API_PATH}/by-name/${encodeURIComponent(
+      projectName
+    )}`,
   });
 }
 
@@ -198,7 +200,9 @@ export async function fetchProjectSettings(
   projectName: string
 ): Promise<ProjectSettings> {
   return http.get(accountId, {
-    url: `${DEVELOPER_PROJECTS_API_PATH}/${encodeURIComponent(projectName)}/settings`,
+    url: `${DEVELOPER_PROJECTS_API_PATH}/${encodeURIComponent(
+      projectName
+    )}/settings`,
   });
 }
 
--- a/lib/path.ts
+++ b/lib/path.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 import unixify from 'unixify';
 import { ALLOWED_EXTENSIONS } from '../constants/extensions';
+import os from 'os';
 
 export function convertToUnixPath(_path: string): string {
   return unixify(path.normalize(_path));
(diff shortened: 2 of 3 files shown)


camden11 (commented):
~Was there a reason for all of the changes in projects.test.ts and projects.ts? Looks like its just indentation changes and nothing else.~

Ah sorry, saw your comment about that. I'd say revert since there are no other changes in those files. They also don't trigger any prettier warnings for me.

We should also probably have a test or two for the `untildify` function

api/projects.ts
```diff
@@ -76,7 +76,9 @@ export async function fetchProject(
   projectName: string
 ): Promise<Project> {
   return http.get(accountId, {
-    url: `${DEVELOPER_PROJECTS_API_PATH}/by-name/${encodeURIComponent(projectName)}`,
+    url: `${DEVELOPER_PROJECTS_API_PATH}/by-name/${encodeURIComponent(
+      projectName
+    )}`,
```
joe-yeager, line 81: Strange, I just pulled `main` and ran `yarn lint` and they seem to be fine on there 🤔

lib/path.ts
```diff
@@ -129,3 +130,15 @@ export function isValidPath(_path: string): boolean {
 
   return true;
 }
+
+// Based on the untildify package: https://github.com/sindresorhus/untildify/blob/main/index.js
+export function untildify(pathWithTilde: string): string {
```
joe-yeager, line 135: This method LGTM, just a +1 to Camden's comment about adding some tests.

lib/__tests__/path.test.ts
```diff
@@ -1,139 +1,241 @@
-import path, { PlatformPath } from 'path';
-import { splitHubSpotPath, splitLocalPath } from '../path';
+import os from 'os';
```
joe-yeager, line 1: 👏 Thanks for adding these!

lib/__tests__/path.test.ts
```diff
@@ -1,139 +1,241 @@
...
+
+    test('does not modify paths without tilde', () => {
+      expect(untildify('/absolute/path/file.txt')).toBe(
+        '/absolute/path/file.txt'
+      );
+    });
+
+    test('throws TypeError for non-string input', () => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
joe-yeager, line 237: This is just a FYI, you don't need to change it.  But using `// @ts-expect-error The reason for the error` shouldn't trigger `eslint` so you can get away with just the one comment

joe-yeager (approved):
LGTM!

lib/__tests__/path.test.ts
kemmerle: Ooh, nice, thanks for the tip!

api/projects.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/188/commits/77b753a1f91c6307b6555f4691d7d1e965e1e36f

PR #189, opened 2024-09-04
Add fetchEnabledFeatures

camden11:
## Description and Context

This adds a new `fetchEnabledFeatures` function to the `localDevAuth` api module that we can use to check gates for an account.

See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/590

## Who to Notify


brandenrodgers (approved):
🚢

PR #190, opened 2024-09-05
fix: Use path.sep to fix path check on Windows

brandenrodgers:
## Description and Context

This fixes an issue on Windows where the path wasn't being checked correctly. The [path.sep](https://nodejs.org/api/path.html#pathsep) value will be `/` on MacOS and `\\` on Windows.

I'm not sure what's going on with Prettier, but it formatted a few other files for some reason


camden11 (commented):
I think the same thing with prettier happened to Allison. Could you try deleting node modules and re-installing and see if that keeps these files the same? Just want to make sure we don't end up in another situation where they go back and forth every PR.

path.sep stuff looks good though

PR #191, opened 2024-09-06
fix: Return the tracking request when not authenticated 

brandenrodgers:
## Description and Context

This is for consistency between the authed and unauthed tracking events. I wasn't able to test this, but I'm fairly certain this is preventing our unauthed requests from sending correctly in some cases. The impact is that our tracking for successful runs of `hs init` isn't accurate


kemmerle (approved):
:shipit:

PR #192, opened 2024-09-17
Refactor config p.2

kemmerle:
## Description and Context
Base branch: https://github.com/HubSpot/hubspot-local-dev-lib/pull/172
Depends on https://github.com/HubSpot/hubspot-cli/pull/1177

This is the second part of the work to refactor the CLI config file. I decided to split the work up to make it easier to review and test. This PR enables the `hs accounts` bucket commands and accounts for the new `defaultAccount` and `accounts` properties in the new config. 

## TODO

- [ ] TEST. We need to test out the `hs auth`, `hs init`, and `hs accounts` commands with the old config, new config, and both configs in place (where we want the new config to override any old configs). In addition, we need to test as many CI commands as possible to ensure that these changes don't impact commands in unexpected ways. 
- [ ] Address any feedback. 

## Who to Notify


camden11 (commented):
Also just wanted to run through everything here before testing, but this is looking good

kemmerle:
SUMMARY OF WORK 9/27:

https://github.com/HubSpot/hubspot-cli/pull/1177#issuecomment-2379491303

config/CLIConfiguration.ts
```diff
@@ -480,7 +489,12 @@ class CLIConfiguration {
     }
 
     if (accountId) {
-      this.addOrUpdateAccount({ accountId, name: newName, env: this.getEnv() });
+      this.addOrUpdateAccount({
+        accountId,
+        name: newName,
+        env: this.getEnv(),
+        accountType: accountConfigToRename.accountType,
```
brandenrodgers, line 496: Why do we need to include the account type here if all we're doing is renaming the existing account? Same question for `env`

config/CLIConfiguration.ts
```diff
@@ -258,8 +258,17 @@ class CLIConfiguration {
...
-        null;
+      return (
+        this.config.accounts.find(account => account.accountId === accountId) ||
+        null
+      );
+    }
+    return null;
+  }
+
+  getConfigAccounts(): Array<CLIAccount_NEW> | null {
```
brandenrodgers, line 269: This can just be `getAccounts()`, which is more inline with the existing `getAccount()` function above

brandenrodgers: I think we could do the same rename for `getConfigAccountIndex()` --> `getAccountIndex()`. These are semi-nits, but I think the consistency will help later on

kemmerle:
Including the account type here addresses a bug with the rename function,
where it was converting all accounts to standard accounts. Specifying the
accounttype and env keeps them from being overwritten.

config/CLIConfiguration.ts
kemmerle: Including the account type here addresses a bug with the rename function, where it was converting all accounts to standard accounts. Specifying the accounttype and env keeps them from being overwritten.

brandenrodgers: huh that's odd. Does that mean `addOrUpdateAccount` has unexpected behavior where it unintentionally overrides config values that aren't specified?

brandenrodgers: If you're just merging this PR into `refactor-config` though, this doesn't have to be a blocker. Just something we should do eventually before merging the parent branch

brandenrodgers (approved):
Just a comment about removing utils from the exports, but otherwise lgtm!

package.json
```diff
@@ -53,10 +53,12 @@
     "./errors/*": "./errors/*.js",
     "./http": "./http/index.js",
     "./http/*": "./http/*.js",
+    "./config/getAccountIdentifier": "./config/getAccountIdentifier.js",
     "./config": "./config/index.js",
     "./constants/*": "./constants/*.js",
     "./models/*": "./models/*.js",
-    "./types/*": "./types/*.d.ts"
+    "./types/*": "./types/*.d.ts",
+    "./utils/*": "./utils/*.js"
```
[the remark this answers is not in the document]
kemmerle: https://github.com/HubSpot/hubspot-local-dev-lib/pull/192/commits/68f22e95fb8d085d0abc3a2257383ef5fc983914

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/192/commits/de06140671bf92e18ce4712bdbcbd2f36fac944b

PR #195, opened 2024-09-25
chore: Add tests for CMS utils

kemmerle:
## Description and Context
Fixes HubSpot CLI issue 492. 

As part of our efforts to expand test coverage, I added tests for:

 `cms/functions.ts`
 `cms/themes.ts`
 `cms/watch.ts`
 `cms/validate.ts`

The original issue also included `cms/processFieldsJs.ts`, but I believe we need to refactor that file before we can write tests for it. Happy to explain in more detail in a Zoom/Huddle, but the tl;dr is that the test suite never runs because the moment it does try to run, it runs the `dynamicImport` function. 

## TODO

- [ ] Address feedback

## Who to Notify

@camden11 @brandenrodgers @joe-yeager


lib/__tests__/validate.test.ts
```diff
@@ -0,0 +1,112 @@
...
+import { validateHubl } from '../../api/validateHubl';
+import { walk } from '../fs';
+import { lint } from '../cms/validate';
+import { LintResult, Validation } from '../../types/HublValidation';
+
+jest.mock('fs-extra');
+jest.mock('../../api/validateHubl');
+jest.mock('../fs');
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kemmerle, line 11: TypeScript was being overly strict and requiring that I mock out `fs.stat` and `fs.readFile` in their entirety. Opted for this solution instead.

lib/__tests__/watch.test.ts
```diff
@@ -0,0 +1,191 @@
...
+import { watch } from '../cms/watch';
+import { MODE } from '../../constants/files';
+
+jest.mock('chokidar');
+jest.mock('axios');
+jest.mock('p-queue');
+jest.mock('../cms/uploadFolder');
+
+describe('watch function', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kemmerle, line 15: TypeScript was being overly strict and requiring that I mock out chokidar in its entirety. Opted for this solution instead.

lib/__tests__/watch.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    chokidarMock = {
+      watch: jest.fn().mockReturnThis(),
+      on: jest.fn().mockReturnThis(),
+    };
+    (chokidar.watch as jest.Mock).mockReturnValue(chokidarMock);
+
+    pQueueAddMock = jest.fn();
+
+    // @ts-expect-error test case
+    PQueue.mockImplementation(() => ({
```
kemmerle, line 29: TypeScript was being overly strict and requiring that I mock out PQueue in its entirety. Opted for this solution instead.

lib/__tests__/functions.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  typeof fs.existsSync
+>;
+const mockedFsReadFileSync = fs.readFileSync as jest.MockedFunction<
+  typeof fs.readFileSync
+>;
+
+describe('functions', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
```
joe-yeager, line 30: Nit: we have [clearMocks set to true](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/jest.config.js#L11), so this should be redundant.

lib/__tests__/themes.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    mockedFindup.mockReturnValue('/path/to/theme.json');
+    mockedGetEnv.mockReturnValue('prod');
+    mockedGetHubSpotWebsiteOrigin.mockReturnValue('https://prod.hubspot.com');
+
+    const result = getThemePreviewUrl('/path/to/file', 12345);
+
+    expect(getEnv).toHaveBeenCalledWith(12345);
+    expect(getHubSpotWebsiteOrigin).toHaveBeenCalledWith(ENVIRONMENTS.PROD);
+    expect(result).toBe(
+      'https://prod.hubspot.com/theme-previewer/12345/edit/to'
```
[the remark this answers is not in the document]
joe-yeager: I'm not following, do you mean you think the path is incorrect?

lib/__tests__/validate.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  html: '',
+};
+
+describe('lint function', () => {
+  const accountId = 123;
+  const filePath = 'test.html';
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
```
joe-yeager, line 58: Same as other nit

lib/__tests__/watch.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      disableInitial: false,
+      notify: '',
+      commandOptions: {},
+      filePaths: [],
+    };
+    const postInitialUploadCallback = jest.fn();
+
+    (uploadFolder as jest.Mock).mockResolvedValueOnce([]);
+
+    await watch(accountId, src, dest, options, postInitialUploadCallback);
```
joe-yeager, line 74: I noticed that sometimes the tests `await watch` and sometimes they don't.  I gave it a quick glance, and it doesn't appear `async`, so maybe we can remove the `await` in those instances.

joe-yeager (approved):
LGTM!  Just a few nits and a couple non-blocking questions

lib/__tests__/watch.test.ts
kemmerle: I wrote an earlier iteration of these tests that was async and then realized it wasn't necessary 💀 Some awaits survived.

lib/__tests__/functions.test.ts
```diff
@@ -0,0 +1,159 @@
...
+const mockedGetCwd = getCwd as jest.MockedFunction<typeof getCwd>;
+const mockedFindup = findup as jest.MockedFunction<typeof findup>;
+const mockedFsExistsSync = fs.existsSync as jest.MockedFunction<
+  typeof fs.existsSync
+>;
+const mockedFsReadFileSync = fs.readFileSync as jest.MockedFunction<
+  typeof fs.readFileSync
+>;
+
+describe('functions', () => {
```
brandenrodgers, line 27: Can you update the top level names of these to match the folder structure? i.e. `lib/cms/functions`?

lib/__tests__/functions.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      );
+
+      // Check that the config file was updated
+      expect(fs.writeFileSync).toHaveBeenCalledWith(
+        '/mock/dest/testFolder.functions/serverless.json',
+        expect.any(String)
+      );
+
+      // Verify that the logger was called with the correct messages
+      jest.spyOn(logger, 'log').mockImplementation(() => {
```
brandenrodgers, line 85: IMO we don't need to have tests for the logged output. That might make these tests a little too brittle

lib/__tests__/themes.test.ts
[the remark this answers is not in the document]
brandenrodgers: I did some digging for this and I think I know what's going on.

This getThemePreviewURL util gets called during `hs upload` [here](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/commands/upload.js#L58). After all of the uploads finish, the command checks to see if the user ran the upload from within a CMS theme folder. It checks by looking for the theme.json file. 

The theme preview path uses the name of the folder that the theme.json file lives in. So if your folder structure looks like this:
```
/ 
..my-theme/
....theme.json
....templates/
....assets/    
```

The `getThemeNameFromPath` function takes the full absolute path to the theme.json file and then returns the folder that the file lives inside of. So it would generate something like `/Users/brodgers/src/my-theme/theme.json` and would return `my-theme` as the path.

The theme preview path would be something like `https://app.hubspot.com/theme-previewer/<account_id>/edit/my-theme`. So your test is working correctly. The test would be easier to follow if you changed the[ mock of findUp](https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/files#diff-514a5be7f3e52cf7cb8d108d7a7be2630da79c7dbfe7dfb428abd04c3c22dc21R26) to return something like `/src/my-theme/theme.json`

Weirdly, this functionality doesn't work for me when I try to test it while running the CLI locally. But you can test it by running `hs project upload` from the root of a CMS theme (just use the default values for each prompt)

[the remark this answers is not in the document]
brandenrodgers: Edit: I think I have a fix for `getThemePreviewUrl` to make it work locally. I can PR it

lib/__tests__/validate.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    theme_field_dependencies: [],
+    template_type_ids: null,
+    exact_path_references: [],
+    required_scopes_to_render: [],
+  },
+  renderingErrors: [],
+  html: '',
+};
+
+describe('lint function', () => {
```
brandenrodgers, line 52: Can you update this to include a wrapper `describe()` that includes the path to this file? The current pattern is:
```
path/to/file
  function 1
  function 2
```

lib/__tests__/watch.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    expect(chokidarMock.on).toHaveBeenCalledWith(
+      'change',
+      expect.any(Function)
+    );
+    const changeCallback = chokidarMock.on.mock.calls[2][1];
+    const filePath = 'changed-file-path.html';
+
+    changeCallback(filePath);
+
+    expect(logger.debug).toHaveBeenCalledWith(
```
brandenrodgers, line 158: Curious to hear what others think about this too, but just leaving the same comment about us not needing to test the log output.

lib/__tests__/watch.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      filePaths: [],
+    };
+
+    watch(accountId, src, dest, options);
+
+    expect(chokidarMock.on).toHaveBeenCalledWith(
+      'unlink',
+      expect.any(Function)
+    );
+    const deleteCallback = chokidarMock.on.mock.calls[2][1];
```
brandenrodgers, line 182: I'm having a little trouble following exactly how this is testing the delete. Can you explain how this is working?

My understanding is that:
1. We're starting up the watch command
2. We're checking to make sure that an `unlink` listener was registered with chokidar
Those make sense to me

3. We're grabbing the mocked chokidar callback func, then manually executing it
4. We're making sure that the delete log shows up
This is the part I'm not following. If we're manually triggering the chokidar functions, then are we actually testing that chokidar is doing what we expect it to do? I'd imagine that it would be more effective for us to use fs or something to actually delete a file from within the watched folder to make sure the function subscribed to chokidar triggers.

Let me know if I'm just misinterpreting this though!

lib/__tests__/validate.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/e98326c0500fa5716cd9765c2558f402dabeec3d

lib/__tests__/functions.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/e98326c0500fa5716cd9765c2558f402dabeec3d

lib/__tests__/validate.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/e98326c0500fa5716cd9765c2558f402dabeec3d

lib/__tests__/functions.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/e98326c0500fa5716cd9765c2558f402dabeec3d

lib/__tests__/watch.test.ts
kemmerle: https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/a54efdd0207005059c9f34db23cfaa33e8dc5de5

lib/__tests__/watch.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/195/commits/a54efdd0207005059c9f34db23cfaa33e8dc5de5

lib/__tests__/watch.test.ts
kemmerle: Ok, ran through this again, and this test does need to be asynchronous.

brandenrodgers (approved):
🚢

joe-yeager (approved):

PR #201, opened 2024-10-29
Add a List Tables to HubDB (Another pull request will be made to the HubSpot CLI as well)

05bmckay:
## Description and Context

Just a quick little function that pulls the published tables from the HubDB.

## TODO

I'll make a pull request to add a `hs hubdb list` command in the CLI as well.

## Who to Notify

@brandenrodgers @joe-yeager @ajlaporte

### DIFF
--- a/api/hubdb.ts
+++ b/api/hubdb.ts
@@ -7,6 +7,7 @@ import {
   Row,
   Schema,
   Table,
+  FetchTablesResponse,
 } from '../types/Hubdb';
 
 const HUBDB_API_PATH = 'cms/v3/hubdb';
--- a/api/hubdb.ts
+++ b/api/hubdb.ts
@@ -11,10 +11,10 @@ import {
 
 const HUBDB_API_PATH = 'cms/v3/hubdb';
 
-export function listTables(
+export function fetchTables(
   accountId: number
-): AxiosPromise<Table> {
-  return http.get<Table>(accountId, {
+): AxiosPromise<FetchTablesResponse> {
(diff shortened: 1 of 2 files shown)


api/hubdb.ts
```diff
@@ -11,6 +11,14 @@ import {
 
 const HUBDB_API_PATH = 'cms/v3/hubdb';
 
+export function listTables(
+  accountId: number
+): AxiosPromise<Table> {
```
brandenrodgers, line 16: The response type here is actually going to look a little different. There's an overview of it [here](https://developers.hubspot.com/beta-docs/reference/api/cms/hubdb#get-%2Fcms%2Fv3%2Fhubdb%2Ftables), but it'll look similar to the existing `FetchRowsResponse` type.

api/hubdb.ts
```diff
@@ -11,6 +11,14 @@ import {
 
 const HUBDB_API_PATH = 'cms/v3/hubdb';
 
+export function listTables(
```
brandenrodgers, line 14: This is a small nit, but would you mind updating this to be `fetchTables` just to align better with our existing naming ?

05bmckay: Good call!

api/hubdb.ts
05bmckay: Sorry you're definitely right, forgot to add that type. Thanks!

05bmckay:
@brandenrodgers made those changes, really appreciate you taking the time for the review.

brandenrodgers (approved):
🎉 thanks Carter! LGTM

PR #202, opened 2024-10-31
Consume is_available_for_new_content value from cli prompt

tenas97:
## Description and Context
Closes HubSpot/hubspot-cli-issues/issues/#344
https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/344

### DIFF
--- a/lib/cms/modules.ts
+++ b/lib/cms/modules.ts
@@ -179,7 +179,7 @@ export async function createModule(
     other_assets: [],
     smart_type: 'NOT_SMART',
     tags: [],
-    is_available_for_new_content: false,
+    is_available_for_new_content: moduleDefinition.availableForNewContent,
   };
 
   const folderName = name.endsWith('.module') ? name : `${name}.module`;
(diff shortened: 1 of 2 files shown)


lib/cms/modules.ts
```diff
@@ -179,7 +179,7 @@ export async function createModule(
     other_assets: [],
     smart_type: 'NOT_SMART',
     tags: [],
-    is_available_for_new_content: false,
+    is_available_for_new_content: moduleDefinition.availableForNewContent,
```
TanyaScales, line 182: For the sake of keeping the same pattern - I would destructure the `availableForNewContent` property from the module definition using the const above already doing so for a [few other things on this line](https://github.com/HubSpot/hubspot-local-dev-lib/pull/202/files#diff-34db6be7e79d62e16dc949471534ee89bdbd4aed6f75d3a7f1deea18d48f2473R164-R169)

TanyaScales (approved):
Other than the one comment here -- this LGTM.

Don't merge this in until we're ready to to do some corresponding work on releasing the update to the CLI. Which will take some coordination with the DPG team

brandenrodgers (approved):
lgtm! 👌

PR #205, opened 2024-11-07
Add build script to LDL

camden11:
## Description and Context
This adds a (very slightly) modified version of the build script from the CLI to local-dev-lib. It also adds a LICENSE file (which we didn't have before for some reason. If there was a good reason for that I can remove) and includes the LICENSE and readme as part of the build output.

Up next: add the release script

## Who to Notify


brandenrodgers (approved):
tested and lgtm!

PR #206, opened 2024-11-08
Add release script

camden11:
## Description and Context
This ports the new release script that was recently added to the CLI to local-dev-lib. The release script here is more or less identical, but doesn't contain the additional confirmation prompt when publishing to latest without a beta release. (Beta releases likely won't be as useful as this is a library, but I didn't see any reason to remove the option).

This also includes the recent change to move the logic for `build` into its own file.

Like the CLI, we'll have to do a bit of work to get things in the proper state:
* Release to the experimental tag
* Release to the beta tag
* Add back in the check for local/beta tag sync (and potentially remove the default experimental version)

## Who to Notify

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -6,7 +6,7 @@ import { confirm } from '@inquirer/prompts';
 
 import { name as packageName, version as localVersion } from '../package.json';
 import { logger, setLogLevel, LOG_LEVEL } from '../lib/logger';
-import { build } from './build';
+import { build } from './lib/build';
 
 const exec = promisify(_exec);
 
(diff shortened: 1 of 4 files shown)


scripts/release.ts
```diff
@@ -0,0 +1,257 @@
...
+    // next: currentNextTag,
+    experimental: currentExperimentalTag,
+  } = await getDistTags();
+
+  // if (!isExperimental && currentNextTag !== localVersion) {
+  //   logger.error(
+  //     `Local package.json version ${localVersion} is out of sync with published version ${currentNextTag}`
+  //   );
+  //   process.exit(EXIT_CODES.ERROR);
+  // }
```
camden11, line 146: Will add this back in after the first beta release

joe-yeager (approved):

brandenrodgers (approved):
🙌

PR #209, opened 2024-11-13
fix: Update port manager shutdown flow

joe-yeager:
## Description and Context
Updates the portManager shutdown flow to call a health check endpoint before attempting to shutdown the server.

## TODO

- [x] Update the CLI code accordingly
- [x] The `hs doctor` PR was making use of the `isPortManagerServerRunning` and will need to be updated to use the `isPortManagerPortAvailable`

### DIFF
--- a/lib/portManager.ts
+++ b/lib/portManager.ts
@@ -1,15 +1,31 @@
 import axios from 'axios';
 
-import { PortManagerServer } from '../utils/PortManagerServer';
-import { detectPort } from '../utils/detectPort';
+import {
+  HEALTH_CHECK_PATH,
+  PortManagerServer,
+  SERVICE_HEALTHY,
+} from '../utils/PortManagerServer';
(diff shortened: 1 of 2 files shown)


lib/portManager.ts
```diff
@@ -1,15 +1,31 @@
...
+import { logger } from './logger';
 
 export const BASE_URL = `http://localhost:${PORT_MANAGER_SERVER_PORT}`;
 
+export async function isPortManagerPortAvailable(): Promise<boolean> {
+  return (
+    (await detectPort(PORT_MANAGER_SERVER_PORT)) === PORT_MANAGER_SERVER_PORT
+  );
+}
+
```
joe-yeager, line 20: This is primarily for use with the `hs doctor` command since it is currently using `isPortManagerServerRunning` to check if the port is available and the logic of that is changing

kemmerle (approved):
LGTM

PR #210, opened 2024-11-14
fix!: Wrap Axios type and export new type

joe-yeager:
## Description and Context
Wraps the `AxiosPromise` in a `HubSpotPromise` type and exports it so that integrators don't need to use `ReturnType` annotations to have access to the type.

### DIFF
--- a/http/index.ts
+++ b/http/index.ts
@@ -16,7 +16,7 @@ import { HubSpotHttpError } from '../models/HubSpotHttpError';
 
 const i18nKey = 'http.index';
 
-export type HubSpotResponse<T = unknown> = Promise<AxiosResponse<T>>;
+export type HubSpotResponse<T = unknown> = AxiosPromise<T>;
 
 axios.interceptors.response.use(undefined, error => {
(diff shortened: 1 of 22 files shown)


http/index.ts
```diff
@@ -16,6 +16,8 @@ import { HubSpotHttpError } from '../models/HubSpotHttpError';
 
 const i18nKey = 'http.index';
 
+export type HubSpotResponse<T = unknown> = AxiosPromise<T>;
```
joe-yeager, line 19: This is the primary change, all of the others are reactionary.

http/index.ts
```diff
@@ -16,6 +16,8 @@ import { HubSpotHttpError } from '../models/HubSpotHttpError';
 
 const i18nKey = 'http.index';
 
+export type HubSpotPromise<T = unknown> = AxiosPromise<T>;
```
camden11, line 19: LDL currently defines all types within the `types` dir, so this should live in `types/http`

camden11 (commented):
Just one minor thing, LGTM otherwise

camden11 (approved):

PR #211, opened 2024-11-18
More release script improvements

camden11:
## Description and Context
This fixes one major bug and makes a few more improvements to the release script
* Previously, the script failed when attempting to update the `next` tag after releases to latest. This is due to a slight difference between how `npm dist-tag` runs within `spawn` and how it runs directly in a terminal. Explicity specifying a `--registry` fixes this.
  * If the `dist-tag` command _does_ fail, the script no longer runs cleanup (since the release itself was successful) and provides instructions for manually updating the `next` tag and pushing to Github.
* Added a confirmation prompt when attempting to release changes to `latest` that haven't already been released to `next`
* Added a prompt for NPM one-time password so you don't have to open your browser. This password seems to work for both the publish and dist-tag operation even if it expires between them, but passing in an expired code will simply open a second prompt so this shouldn't be an issue.
* Opens the browser to remind us to create a new release on Github after all successful runs

## Who to Notify

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "0.0.1-experimental.2",
+  "version": "0.0.1-experimental.3",
   "description": "Provides library functionality for HubSpot local development tooling, including the HubSpot CLI",
   "main": "lib/index.js",
   "repository": {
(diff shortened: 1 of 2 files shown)


joe-yeager (approved):
LGTM!

PR #216, opened 2024-11-22
chore!: Rename Mode to CmsPublishMode

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/718
This updates all references to `mode` to `cmsPublishMode` and renames all related functions and variables. It also changes the `defaultMode` setting within `hubspot.config.yml` to `defaultCmsPublishMode`. Both the new and legacy config modules have been updated to be able to read configs with either `defaultMode` or `defaultCmsPublishMode`, but will now always  use `defaultCmsPublishMode` when making updates to config files.

Currently released on `v0.1.0-experimental.0`
## TODO
Test thoroughly. Also might be worth updating the config unit tests to test the backwards compatibility functionality/

## Who to Notify

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "2.3.0",
+  "version": "0.1.0-experimental.0",
   "description": "Provides library functionality for HubSpot local development tooling, including the HubSpot CLI",
   "main": "lib/index.js",
   "repository": {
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "2.3.0",
+  "version": "0.1.0-experimental.0",
```
kemmerle, line 3: Remember to change this back when you merge.

kemmerle (approved):
Tested out the changes with both the new and deprecated config, updating both `defaultMode` and `defaultCmsPublishMode.` It all appears to be WAD to me, unless there are other test scenarios I'm missing.

PR #221, opened 2024-12-16
Chore: update project types

camden11:
## Description and Context
This fixes a few inaccuracies with types related to projects + removes the `OptionalError` types which is no longer needed.

Type changes based on API responses herer: https://tools.hubteam.com/api-catalog/services/DFSBuildPipelineService/v1/spec/internal

## Who to Notify

### DIFF
--- a/api/projects.ts
+++ b/api/projects.ts
@@ -7,6 +7,7 @@ import {
   UploadProjectResponse,
   ProjectSettings,
   FetchPlatformVersionResponse,
+  WarnLogsResponse,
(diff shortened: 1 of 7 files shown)


brandenrodgers (approved):
lgtm. This will require a major version bump correct?

camden11:
Good question. I wasn't really thinking of it as a breaking change since it only changes types and not functionality. But it _does_ break types. I'll look into what other people are doing for things like this

PR #223, opened 2024-12-17
Fix auth type for deprecated config

brandenrodgers:
## Description and Context

I missed the auth type in the deprecated config in https://github.com/HubSpot/hubspot-local-dev-lib/pull/220. We also need to fix it here.


kemmerle (approved):
:shipit:

PR #224, opened 2025-01-10
fix: Move GitHub downloads to optimized method

joe-yeager:
## Description and Context

Changes made:
- Deprecate `downloadGithubRepoContents` because it is inefficient and tends to hit rate limits
- Replace `downloadGithubRepoContents` with `cloneGithubRepo` to optimize the number of calls to GitHub
- Make any required tweaks to account for the transition

### DIFF
--- a/lib/__tests__/functions.test.ts
+++ b/lib/__tests__/functions.test.ts
@@ -1,7 +1,7 @@
 import fs, { PathLike } from 'fs-extra';
 import findup from 'findup-sync';
 import { getCwd } from '../path';
-import { downloadGithubRepoContents } from '../github';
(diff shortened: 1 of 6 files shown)


brandenrodgers:
I think the answer is yes, but just to confirm. Is this backwards compatible?

joe-yeager:
> I think the answer is yes, but just to confirm. Is this backwards compatible?

Absolutely!  [The companion PR in the CLI](https://github.com/HubSpot/hubspot-cli/pull/1326) outlines the testing steps to verify that is the case

brandenrodgers (approved):
This lgtm 👍 lets get it out on a LDL beta release next week so we can test more easily over in the CLI

PR #228, opened 2025-01-17
Feature: Add default account overrides with `.hsaccount` file

kemmerle:
## Description and Context
In this PR, I've implemented support for default account overrides with the new centralized configuration file. 

Now, customers can place a `.hs-account` file in any directory where they would like the default account overridden. The `.hs-account` file will contain a single numerical entry: `1234567`, or the account ID. Any changes made in the directory and its subdirectories will apply in the account stipulated in the `.hs-account` file. This will work very similarly to our nest configs currently do. 

The advantages are:

1) We will still have one centralized configuration file and therefore one single source of truth. 
2) We will create a command to auto-generate the `.hs-account` file, so that customers need never have to manually interact with the config. 
3) It's much more secure, if a customer accidentally uploads the file to GitHub--it contains no sensitive information besides an account ID. 

## TODO

- [ ] Address feedback 

## Who to Notify


lang/en.json
```diff
@@ -263,6 +263,12 @@
           "invalidInput": "A 'defaultAccount' with value of number or string is required to update the config."
         }
       },
+      "getResolvedDefaultAccountForCWD": {
+        "errors": {
+          "invalidAccountId": "In the default override file (.hs-account), the account ID is required to be a number. Please delete the current file and generate it using `hs accounts create-override`",
```
kemmerle, line 268: I will implement `uiCommandReference` and other styling with these errors, but I wanted to get feedback on the copy before I do. Will ping Jono to take a look.

brandenrodgers: I don't think we want to reference any CLI commands from within this library. This lib should be mostly agnostic of the CLI commands b/c it can be used in other places like the VS code extension. Instead we can throw a reasonable error from here, catch it in the CLI, and then log out any sort of more-specific errors from there.

brandenrodgers:
The code in here seems reasonable to me! I'll test this out early next week 🎉

config/CLIConfiguration.ts
```diff
@@ -228,23 +235,50 @@ class _CLIConfiguration {
...
+    const source = fs.readFileSync(defaultOverrideFile, 'utf8');
+    const accountId = Number(source);
+
+    if (isNaN(accountId)) {
+      logger.error(
+        i18n(
+          `${i18nKey}.getResolvedDefaultAccountForCWD.errors.invalidAccountId`
+        )
+      );
+      process.exit(1);
```
joe-yeager, line 261: I think we should replace the `process.exit` calls in this file with an exception, we don't want to assume that these are always critical errors and should give the consuming application the chance to make that decision.

config/CLIConfiguration.ts
```diff
@@ -228,23 +235,50 @@ class _CLIConfiguration {
...
+    }
+
+    if (this.config && this.config.accounts) {
```
joe-yeager, line 264: Under what circumstances are `this.config` and `this.config.accounts` undefined?  That feels like it should possibly be an exception and not returning the value in `accountId`

kemmerle: This conditional is to avoid a TypeScript error. In reality, if either `this.config` or `this.config.accounts` are undefined, we will throw an error much earlier. 

We also don't return `accountId` in this conditional--its purpose is only to ensure that the provided `accountId` from the `.hs-account` file is listed in the `this.config.accounts` array. Given that info, do I still need to make changes?

joe-yeager: > This conditional is to avoid a TypeScript error. In reality, if either this.config or this.config.accounts are undefined, we will throw an error much earlier.
I don't think we should count on that behavior and this function should be self contained and still assume there is a chance that `this.config` and `this.config.account` are nullable. This `if` statement modifies the behavior so that if the opposite of this conditional is true (`!this.config || !this.config.accounts`), we will skip the `if(!account)` check and the log statement and just directly return the `accountId` on line 281.

I think we should either remove this `if` statement and change the `find` call to something like this:
```javascript
      const account = this.config?.accounts?.find(
        account => account.accountId === accountId
      );
```

Or I think we should move the `if (!account)` block out of the `if (this.config && this.config.accounts)`

kemmerle: Thanks for the explanation, Joe! Always great to learn from you!

kemmerle: > This if statement modifies the behavior so that if the opposite of this conditional is true (!this.config || !this.config.accounts), we will skip the if(!account) check and the log statement and just directly return the accountId on line 281.

This makes me want to go back and check some of my work now 👀

kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/2132b2299444bdcd8c31c37b4f6bd53cf28295b4

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/2132b2299444bdcd8c31c37b4f6bd53cf28295b4

lang/en.json
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/2132b2299444bdcd8c31c37b4f6bd53cf28295b4

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,53 @@ class _CLIConfiguration {
...
+      this.getResolvedDefaultAccountForCWD() ||
+      this.config?.defaultAccount ||
+      null
+    );
+  }
+
+  getDefaultAccountOverrideFilePath(): string | null {
+    return findup([DEFAULT_ACCOUNT_OVERRIDE_FILE_NAME], {
+      cwd: getCwd(),
+    });
```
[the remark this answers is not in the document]
joe-yeager: Discussed in person

joe-yeager (approved):

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,53 @@ class _CLIConfiguration {
...
-  getResolvedDefaultAccountForCWD(
-    nameOrId: string | number
-  ): CLIAccount_NEW | null {
-    return this.getAccount(nameOrId);
+  getResolvedDefaultAccountForCWD(): number | null {
+    const defaultOverrideFile = this.getDefaultAccountOverrideFilePath();
+    if (!defaultOverrideFile) {
+      return null;
+    }
+    const source = fs.readFileSync(defaultOverrideFile, 'utf8');
```
brandenrodgers, line 258: This `readFileSync` call can fail (for example, a permission denied error) so we'll probably want a try catch around it to log out a FileSystemError.

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,53 @@ class _CLIConfiguration {
...
+    const accountId = Number(source);
```
brandenrodgers, line 259: What does `Number` spit out if you pass it something that's not a number string?

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,53 @@ class _CLIConfiguration {
   }
 
   getDefaultAccount(): string | number | null {
-    return this.config && this.config.defaultAccount
-      ? this.config.defaultAccount
-      : null;
+    return (
+      this.getResolvedDefaultAccountForCWD() ||
```
brandenrodgers, line 241: Should `getResolvedDefaultAccountForCWD` return the configured account name for the account instead of the accountId? It would make this more consistent with the previous return type for `getDefaultAccount`, which technically can support the account ID but typically is the account name string.

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,53 @@ class _CLIConfiguration {
...
-  //  - /src/brodgers/customer-project-1: customer-account1
-  //  - /src/brodgers/customer-project-2: customer-account2
-  // "/src/brodgers/customer-project-1" is the path to the project dir
-  // "customer-account1" is the name of the account to use as the default for the specified dir
-  // These defaults take precedence over the standard default account specified in the config
-  getResolvedDefaultAccountForCWD(
-    nameOrId: string | number
-  ): CLIAccount_NEW | null {
-    return this.getAccount(nameOrId);
+  getResolvedDefaultAccountForCWD(): number | null {
```
brandenrodgers, line 253: nit, this naming `getResolvedDefaultAccountForCWD` made a little more sense when we were planning to implement the previous idea. 

Maybe something like `findDefaultAccountOverrideFile()` and `getCWDAccountOverride()`?

config/CLIConfiguration.ts
kemmerle: It returns `NaN`--that's why I've included the `isNaN` check on line 261, so we can throw an error if the accountId cannot be converted to a number.

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/175f2106db5c2a7d37a17330d49148fa274a06a1

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/175f2106db5c2a7d37a17330d49148fa274a06a1

config/CLIConfiguration.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/175f2106db5c2a7d37a17330d49148fa274a06a1

config/CLIConfiguration.ts
```diff
@@ -228,23 +237,66 @@ class _CLIConfiguration {
...
+      if (isNaN(accountId)) {
+        throw new Error(
+          i18n(`${i18nKey}.getCWDAccountOverride.errorHeader`, {
+            hsAccountFile: defaultOverrideFile,
+          }),
+          {
+            cause: DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID,
+          }
+        );
+      }
```
joe-yeager, line 267: Throwing inside in a `try` block is an anti-pattern because it essentially acts as a `goto` to the corresponding `catch` block.  It seems like we could just wrap this line in the `try` block
```
const source = fs.readFileSync(defaultOverrideFile, 'utf8');
```

and move the these two if statements outside of the `try` block entirely and throw the errors to be handled by the invoking method.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "3.1.2",
+  "version": "0.1.2-experimental.0",
```
joe-yeager, line 3: This seems unintentional.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-local-dev-lib/pull/228/commits/460154dc57bf41d48bb20e6ba91f3f27bfed83ad

joe-yeager (approved):

PR #229, opened 2025-01-27
chore: Pin deps

joe-yeager:
## Description and Context

Pin all production dependencies in the library.  Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/229


camden11 (approved):

camden11 (approved):

.yarnrc
```diff
@@ -0,0 +1 @@
+--install.no-lockfile true
```
joe-yeager, line 1: This will prevent the creation/use of a lockfile

PR #230, opened 2025-01-27
feat: Update deployProject to support v3

joe-yeager:
## Description and Context

Add a path to conditionally perform the v3 upload


brandenrodgers (approved):
Thanks for adding tests 🙏

PR #231, opened 2025-02-03
Dynamic config

camden11:

### DIFF
--- a/constants/config.ts
+++ b/constants/config.ts
@@ -22,3 +22,7 @@ export const HUBSPOT_ACCOUNT_TYPE_STRINGS = {
   APP_DEVELOPER: i18n('lib.accountTypes.appDeveloper'),
   STANDARD: i18n('lib.accountTypes.standard'),
 } as const;
+
+export const CONFIG_FLAGS = {
+  USE_CUSTOM_OBJECT_HUBFILE: 'useCustomObjectHubfile',
+} as const;
--- a/types/Accounts.ts
+++ b/types/Accounts.ts
@@ -18,8 +18,8 @@ export interface HubSpotConfigAccount {
     clientSecret?: string;
     scopes?: Array<string>;
   };
-  sandboxAccountType?: string | null;
-  parentAccountId?: number | null;
+  sandboxAccountType?: string;
+  parentAccountId?: number;
   apiKey?: string;
   personalAccessKey?: string;
 }
--- a/types/Accounts.ts
+++ b/types/Accounts.ts
@@ -16,6 +16,7 @@ export interface HubSpotConfigAccount {
     tokenInfo?: TokenInfo;
     clientId?: string;
     clientSecret?: string;
+    scopes?: Array<string>;
   };
   sandboxAccountType?: string | null;
   parentAccountId?: number | null;
--- a/types/Accounts.ts
+++ b/types/Accounts.ts
@@ -5,7 +5,7 @@ import { ValueOf } from './Utils';
 
 export type AuthType = 'personalaccesskey' | 'apikey' | 'oauth2';
 
-export interface CLIAccount_NEW {
+export interface HubSpotConfigAccount {
   name?: string;
   accountId: number;
   accountType?: AccountType;
@@ -23,25 +23,9 @@ export interface CLIAccount_NEW {
   personalAccessKey?: string;
 }
 
-export interface CLIAccount_DEPRECATED {
-  name?: string;
-  portalId?: number;
-  defaultCmsPublishMode?: CmsPublishMode;
-  env: Environment;
-  accountType?: AccountType;
-  authType?: AuthType;
-  auth?: {
-    tokenInfo?: TokenInfo;
-    clientId?: string;
-    clientSecret?: string;
-  };
-  sandboxAccountType?: string | null;
-  parentAccountId?: number | null;
-  apiKey?: string;
-  personalAccessKey?: string;
-}
-
-export type CLIAccount = CLIAccount_NEW | CLIAccount_DEPRECATED;
+export type DeprecatedHubSpotConfigAccountFields = {
+  portalId: number;
+};
 
 export type GenericAccount = {
   portalId?: number;
@@ -56,22 +40,12 @@ export type TokenInfo = {
   refreshToken?: string;
 };
 
-export interface PersonalAccessKeyAccount_NEW extends CLIAccount_NEW {
+export interface PersonalAccessKeyConfigAccount extends HubSpotConfigAccount {
   authType: 'personalaccesskey';
   personalAccessKey: string;
 }
 
-export interface PersonalAccessKeyAccount_DEPRECATED
-  extends CLIAccount_DEPRECATED {
-  authType: 'personalaccesskey';
-  personalAccessKey: string;
-}
-
-export type PersonalAccessKeyAccount =
-  | PersonalAccessKeyAccount_NEW
-  | PersonalAccessKeyAccount_DEPRECATED;
-
-export interface OAuthAccount_NEW extends CLIAccount_NEW {
+export interface OAuthConfigAccount extends HubSpotConfigAccount {
   authType: 'oauth2';
   auth: {
     clientId?: string;
@@ -81,31 +55,12 @@ export interface OAuthAccount_NEW extends CLIAccount_NEW {
   };
 }
 
-export interface OAuthAccount_DEPRECATED extends CLIAccount_DEPRECATED {
-  authType: 'oauth2';
-  auth: {
-    clientId?: string;
-    clientSecret?: string;
-    scopes?: Array<string>;
-    tokenInfo?: TokenInfo;
-  };
-}
-
-export type OAuthAccount = OAuthAccount_NEW | OAuthAccount_DEPRECATED;
-
-export interface APIKeyAccount_NEW extends CLIAccount_NEW {
+export interface APIKeyConfigAccount extends HubSpotConfigAccount {
   authType: 'apikey';
   apiKey: string;
 }
 
-export interface APIKeyAccount_DEPRECATED extends CLIAccount_DEPRECATED {
-  authType: 'apikey';
-  apiKey: string;
-}
-
-export type APIKeyAccount = APIKeyAccount_NEW | APIKeyAccount_DEPRECATED;
-
-export interface FlatAccountFields_NEW extends CLIAccount_NEW {
+export interface HubSpotConfigAccountOptions extends HubSpotConfigAccount {
   tokenInfo?: TokenInfo;
   clientId?: string;
   clientSecret?: string;
@@ -113,19 +68,6 @@ export interface FlatAccountFields_NEW extends CLIAccount_NEW {
   apiKey?: string;
   personalAccessKey?: string;
 }
-
-export interface FlatAccountFields_DEPRECATED extends CLIAccount_DEPRECATED {
-  tokenInfo?: TokenInfo;
-  clientId?: string;
-  clientSecret?: string;
-  scopes?: Array<string>;
-  apiKey?: string;
-  personalAccessKey?: string;
-}
-
-export type FlatAccountFields =
-  | FlatAccountFields_NEW
-  | FlatAccountFields_DEPRECATED;
 
 export type ScopeData = {
   portalScopesInGroup: Array<string>;
@@ -148,10 +90,10 @@ export type EnabledFeaturesResponse = {
   enabledFeatures: { [key: string]: boolean };
 };
 
-export type UpdateAccountConfigOptions =
-  Partial<FlatAccountFields_DEPRECATED> & {
-    environment?: Environment;
-  };
+// export type UpdateAccountConfigOptions =
+//   Partial<FlatAccountFields_DEPRECATED> & {
+//     environment?: Environment;
+//   };
 
 export type PersonalAccessKeyOptions = {
   accountId: number;
--- a/config/config.ts
+++ b/config/config.ts
@@ -1,0 +1,94 @@
+import fs from 'fs';
+import findup from 'findup-sync';
+
+import { DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME } from '../constants/config';
+import { HubSpotConfigAccount } from '../types/Accounts';
+import { HubSpotConfig, Environment, ConfigFlag } from '../types/Config';
+import { CmsPublishMode } from '../types/Files';
+import { logger } from '../lib/logger';
+import { i18n } from '../utils/lang';
+import {
+  getGlobalConfigFilePath,
+  readConfigFile,
+  parseConfig,
+} from './configUtils';
+
+export function getConfigFilePath(): string {
+  const globalConfigFilePath = getGlobalConfigFilePath();
+
+  if (fs.existsSync(globalConfigFilePath)) {
+    return globalConfigFilePath;
+  }
+
+  const localConfigPath = findup([
+    DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME,
+    DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME.replace('.yml', '.yaml'),
+  ]);
+
+  if (!localConfigPath) {
+    throw new Error('@TODO');
+  }
+
+  return localConfigPath;
+}
+
+export function getConfig(
+  configFilePath?: string,
+  useEnv = false
+): HubSpotConfig {
+  if (configFilePath && useEnv) {
+    throw new Error('@TODO');
+  }
+
+  if (useEnv) {
+    // do something else
+  }
+
+  const pathToRead = configFilePath || getConfigFilePath();
+
+  logger.debug(`@TODOReading config from ${pathToRead}`);
+  const configFileSource = readConfigFile(pathToRead);
+
+  return parseConfig(configFileSource);
+}
+
+function getConfigIfExists(): HubSpotConfig | undefined {}
+
+function isConfigValid(config: HubSpotConfig): boolean {}
+
+function createEmptyConfigFile(): void {}
+
+function deleteConfigFile(): void {}
+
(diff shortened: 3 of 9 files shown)


config/README.md
```diff
@@ -10,17 +10,20 @@ The config file is named `huspot.config.yml`.
...
 
-#### getAccountConfig()
+Safely writes updated values to the `hubspot.config.yml` file.
 
-Returns config data for a specific account, given the account's ID.
+#### getConfigAccountById() and getConfigAccountByName()
+
+Returns config data for a specific account, given the account's ID or name.
 
 ## Example config
```
camden11, line 28: I updated the existing content of the readme to reflect the changes I made, but it may make sense to rethink it completely

config/index.ts
```diff
@@ -1,268 +1,406 @@
-import * as config_DEPRECATED from './config_DEPRECATED';
```
camden11, line 1: This file is more or less net new, but because it has the same name as the old one, GH is showing a diff.

config/index.ts
```diff
@@ -1,268 +1,406 @@
...
+  return localConfigFilePath;
 }
 
-export function deleteEmptyConfigFile() {
-  if (CLIConfiguration.isActive()) {
-    return CLIConfiguration.delete();
-  }
-  return config_DEPRECATED.deleteEmptyConfigFile();
+export function getConfigFilePath(): string {
+  const { configFilePathFromEnvironment } = getConfigPathEnvironmentVariables();
```
camden11, line 50: `getConfigFilePath` checks for a `HUBSPOT_CONFIG_PATH` environment variable if the user is using a custom path - this variable will need to be set by the CLI if the user passes in the `--config` flag

config/index.ts
```diff
@@ -1,268 +1,406 @@
...
+
+  return configFilePathFromEnvironment || getDefaultConfigFilePath();
 }
 
-export function getConfig(): CLIConfig | null {
-  if (CLIConfiguration.isActive()) {
-    return CLIConfiguration.config;
+export function getConfig(): HubSpotConfig {
+  const { useEnvironmentConfig } = getConfigPathEnvironmentVariables();
```
camden11, line 56: Same with environment config - a user passes in the `--use-env` flag to the CLI, the CLI should set the `USE_ENVIRONMENT_CONFIG` env variable to tell LDL to use environment variables. These new environment variables let us avoid needing to have configPath and useEnv arguments for every function that accesses the config

config/index.ts
```diff
@@ -1,268 +1,406 @@
...
-export function getAccountId(nameOrId?: string | number): number | null {
-  if (CLIConfiguration.isActive()) {
-    return CLIConfiguration.getAccountId(nameOrId);
+export function getConfigAccountEnvironment(
+  identifier?: number | string
+): Environment {
+  if (identifier) {
+    const config = getConfig();
+
+    const account = getConfigAccountByInferredIdentifier(
```
camden11, line 202: This function allows you to get accounts by name _or_ id (primarily for backwards compatiblity with how the CLI currently works), but opted to use it as minimally as possible to reduce likelihood of bugs

config/index.ts
```diff
@@ -1,268 +1,406 @@
...
+  writeConfigFile(config, getConfigFilePath());
 }
 
-export function updateHttpTimeout(timeout: string): void {
-  if (CLIConfiguration.isActive()) {
-    CLIConfiguration.updateHttpTimeout(timeout);
-  } else {
-    config_DEPRECATED.updateHttpTimeout(timeout);
+export function updateConfigAccount(
+  updatedAccount: HubSpotConfigAccount
```
camden11, line 242: This function now accepts a full updated account rather than just fields to update. This makes it easier to use TS to confirm that updates are valid (for example, TS would have no way to know the auth type of the account you're trying to update, and therefore would allow updating accounts to have invalid auth)

config/index.ts
```diff
@@ -1,268 +1,405 @@
...
-    portalId: accountIdentifier,
-  });
-}
 
-export function updateDefaultAccount(nameOrId: string | number): void {
-  if (CLIConfiguration.isActive()) {
-    CLIConfiguration.updateDefaultAccount(nameOrId);
-  } else {
-    config_DEPRECATED.updateDefaultAccount(nameOrId);
+  const account = getConfigAccountByInferredIdentifier(
```
camden11, line 173: This function allows you to get accounts by name or id (primarily for backwards compatibility with how the CLI currently works), but opted to use it as minimally as possible to reduce likelihood of bugs.

We now only _set_ default accounts using their `accountId`, but this is set up to maintain backwards compatibility with default accounts specified by `name`

config/index.ts
```diff
@@ -1,268 +1,405 @@
...
+
+  if (!account) {
+    throw new Error(
+      i18n('config.setConfigAccountAsDefault.accountNotFound', {
+        accountId: identifier,
+      })
+    );
   }
+
+  config.defaultAccount = account.accountId;
```
camden11, line 283: For consistency, we now always use `accountId` to refer to default accounts

config/index.ts
```diff
@@ -1,268 +1,405 @@
...
-export function getAccountType(
-  accountType?: AccountType,
-  sandboxAccountType?: string | null
-): AccountType {
-  if (CLIConfiguration.isActive()) {
-    return CLIConfiguration.getAccountType(accountType, sandboxAccountType);
+  config.accounts.splice(index, 1);
+
+  if (config.defaultAccount === accountId) {
+    delete config.defaultAccount;
```
camden11, line 342: need to double check but I believe detecting this and prompting you to set a new default is handled by the CLI

config/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+    if (account.authType === OAUTH_AUTH_METHOD.value) {
+      Object.keys(account.auth).forEach(k => {
+        const key = k as keyof OAuthConfigAccount['auth'];
+        if (account.auth?.[key] === undefined) {
+          delete account.auth?.[key];
+        }
+      });
+    }
+
+    if (
```
camden11, line 116: Going through these fields manually makes this function a bit more verbose than it needs to be, but keeps things type safe

config/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+      return HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX;
+    }
+    if (sandboxAccountType.toUpperCase() === 'STANDARD') {
+      return HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX;
+    }
+  }
+  return HUBSPOT_ACCOUNT_TYPES.STANDARD;
+}
+
+export function normalizeParsedConfig(
```
camden11, line 199: This function reads both old and new config formats and converts them to an object with the same shape that can be used around LDL and the CLI

(for example, converts `portals` to `accounts` among other things)

config/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+    process.env[ENVIRONMENT_VARIABLES.HUBSPOT_ACCOUNT_ID] ||
+    process.env[ENVIRONMENT_VARIABLES.HUBSPOT_PORTAL_ID];
+  const refreshToken = process.env[ENVIRONMENT_VARIABLES.HUBSPOT_REFRESH_TOKEN];
+  const hubspotEnvironment =
+    process.env[ENVIRONMENT_VARIABLES.HUBSPOT_ENVIRONMENT];
+  const httpTimeoutVar = process.env[ENVIRONMENT_VARIABLES.HTTP_TIMEOUT];
+  const httpUseLocalhostVar =
+    process.env[ENVIRONMENT_VARIABLES.HTTP_USE_LOCALHOST];
+  const allowUsageTrackingVar =
+    process.env[ENVIRONMENT_VARIABLES.ALLOW_USAGE_TRACKING];
```
camden11, line 268: We now support new env variables for `HTTP_TIMEOUT`, `ALLOW_USAGE_TRACKING` and other things that config files supported but environment configs didn't

constants/environments.ts
```diff
@@ -2,15 +2,3 @@ export const ENVIRONMENTS = {
...
-  HUBSPOT_API_KEY: 'HUBSPOT_API_KEY',
-  HUBSPOT_CLIENT_ID: 'HUBSPOT_CLIENT_ID',
-  HUBSPOT_CLIENT_SECRET: 'HUBSPOT_CLIENT_SECRET',
-  HUBSPOT_PERSONAL_ACCESS_KEY: 'HUBSPOT_PERSONAL_ACCESS_KEY',
-  HUBSPOT_ACCOUNT_ID: 'HUBSPOT_ACCOUNT_ID',
-  HUBSPOT_PORTAL_ID: 'HUBSPOT_PORTAL_ID',
-  HUBSPOT_REFRESH_TOKEN: 'HUBSPOT_REFRESH_TOKEN',
-  HUBSPOT_ENVIRONMENT: 'HUBSPOT_ENVIRONMENT',
-  HTTP_TIMEOUT: 'HTTP_TIMEOUT',
-} as const;
```
camden11, line 16: Moved these to `constants/config` - this file is for _HubSpot_ environtments

types/Accounts.ts
```diff
@@ -56,76 +32,34 @@ export type TokenInfo = {
...
-  clientSecret?: string;
-  scopes?: Array<string>;
-  apiKey?: string;
-  personalAccessKey?: string;
-}
-
-export type FlatAccountFields =
-  | FlatAccountFields_NEW
-  | FlatAccountFields_DEPRECATED;
+export type HubSpotConfigAccount =
```
camden11, line 59: This a) lets us confirm that accounts returned by config methods have valid auth and b) lets us use the `accountType` field to determine exactly what type an account is

types/Accounts.ts
```diff
@@ -148,32 +82,6 @@ export type EnabledFeaturesResponse = {
...
-  scopes: Array<string>;
-  env: Environment;
-};
-
-export type APIKeyOptions = {
-  accountId: number;
-  apiKey: string;
-  env: Environment;
-};
-
```
camden11, line 176: Anywhere that previously used these types now just use the accounts themselves to keep things simple

config/README.md
brandenrodgers: Yeah good call. That could always be done as a follow up

http/index.ts
```diff
@@ -99,17 +96,27 @@ async function withAuth(
...
-  const { params } = axiosConfig;
 
-  return {
-    ...axiosConfig,
-    params: {
-      ...params,
-      hapikey: apiKey,
-    },
-  };
+  if (authType === 'apikey') {
```
brandenrodgers, line 102: Do we have these as constants anywhere?

types/Accounts.ts
```diff
@@ -56,76 +32,34 @@ export type TokenInfo = {
...
-  personalAccessKey?: string;
-}
-
-export type FlatAccountFields =
-  | FlatAccountFields_NEW
-  | FlatAccountFields_DEPRECATED;
+export type HubSpotConfigAccount =
+  | PersonalAccessKeyConfigAccount
+  | OAuthConfigAccount
+  | APIKeyConfigAccount;
```
brandenrodgers, line 62: We can double check, but I'm fairly certain newer versions of the CLI don't support apikey auth anymore. So it might be safe to remove support for it in the new version of this lib

constants/config.ts
```diff
@@ -22,3 +22,24 @@ export const HUBSPOT_ACCOUNT_TYPE_STRINGS = {
...
+  HUBSPOT_PERSONAL_ACCESS_KEY: 'HUBSPOT_PERSONAL_ACCESS_KEY',
+  HUBSPOT_ACCOUNT_ID: 'HUBSPOT_ACCOUNT_ID',
+  HUBSPOT_PORTAL_ID: 'HUBSPOT_PORTAL_ID',
+  HUBSPOT_REFRESH_TOKEN: 'HUBSPOT_REFRESH_TOKEN',
+  HUBSPOT_ENVIRONMENT: 'HUBSPOT_ENVIRONMENT',
+  HTTP_TIMEOUT: 'HTTP_TIMEOUT',
+  HTTP_USE_LOCALHOST: 'HTTP_USE_LOCALHOST',
+  ALLOW_USAGE_TRACKING: 'ALLOW_USAGE_TRACKING',
+  DEFAULT_CMS_PUBLISH_MODE: 'DEFUALT_CMS_PUBLISH_MODE',
+  USE_ENVIRONMENT_CONFIG: 'USE_ENVIRONMENT_CONFIG',
```
brandenrodgers, line 43: Slight nit, but does it make sense for `USE_ENVIRONMENT_CONFIG` to include "hubspot" in the name somewhere? It doesn't matter a ton if the CLI is setting it, but I imagine people can also manually set this if they want to.

config/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+  accountIdentifier: string | number
+): HubSpotConfigAccount | undefined {
+  const { identifier, identifierType } =
+    getAccountIdentifierAndType(accountIdentifier);
+  return accounts.find(account => account[identifierType] === identifier);
+}
+
+export function getConfigAccountIndexById(
+  accounts: Array<HubSpotConfigAccount>,
+  id: string | number
```
brandenrodgers, line 380: Is it possible for accountId to be anything except for a number in this new format?

config/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+): { identifier: string | number; identifierType: 'name' | 'accountId' } {
+  const identifierAsNumber =
+    typeof accountIdentifier === 'number'
+      ? accountIdentifier
+      : parseInt(accountIdentifier);
+  const isId = !isNaN(identifierAsNumber);
+
+  return {
+    identifier: isId ? identifierAsNumber : accountIdentifier,
+    identifierType: isId ? 'accountId' : 'name',
```
brandenrodgers, line 357: Not a blocker, but could it potentially prevent some future pain by making these identifiers constants?

config/__tests__/utils.test.ts
```diff
@@ -0,0 +1,394 @@
...
+      );
+
+      expect(withFieldsRemoved).toEqual(PAK_ACCOUNT);
+    });
+  });
+
+  describe('writeConfigFile()', () => {
+    it('writes formatted config to file', () => {
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      mockFs.ensureFileSync.mockImplementation(() => {});
```
brandenrodgers, line 225: Not a blocker, but you could get rid of the ignores if you make these return undefined instead

config/defaultAccountOverride.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { getCwd } from '../lib/path';
+import {
+  DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID,
+  DEFAULT_ACCOUNT_OVERRIDE_ERROR_ACCOUNT_NOT_FOUND,
+  DEFAULT_ACCOUNT_OVERRIDE_FILE_NAME,
+} from '../constants/config';
+import { i18n } from '../utils/lang';
+import { FileSystemError } from '../models/FileSystemError';
+import { getAllConfigAccounts } from './index';
```
camden11, line 12: I opted to move this stuff to its own file because it doesn't directly interact with the config file

http/index.ts
camden11: Good call! We sure do

config/index.ts
```diff
@@ -1,268 +1,405 @@
...
-  }
-  return config_DEPRECATED.validateConfig();
-}
+function getDefaultConfigFilePath(): string {
+  const globalConfigFilePath = getGlobalConfigFilePath();
 
-export function loadConfigFromEnvironment(): boolean {
-  if (CLIConfiguration.isActive()) {
-    return CLIConfiguration.useEnvConfig;
+  if (fs.existsSync(globalConfigFilePath)) {
```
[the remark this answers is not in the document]
camden11: We're definitely going to need to add some try/catch blocks in places we didn't have them previously, though I think we'll also be able to remove a lot of error throwing from the CLI too (for example, theres a few places where _we_ know accountID exists but TS doesn't so we have to throw if it doesn't).

In this particular case though, `fs.existsSync` doesn't throw - it'll just return false if these nothing at the globalConfigFilePath

config/utils.ts
camden11: Good catch, this should just be number

joe-yeager (approved):
I like the new approach a lot!

brandenrodgers:
Code lgtm 👌 we should chat about what the rollout strategy looks like

config/index.ts
```diff
@@ -1,282 +1,426 @@
...
-    return CLIConfiguration.addOrUpdateAccount({
-      ...configOptions,
-      accountId: accountIdentifier,
-    });
-  }
-  return config_DEPRECATED.updateAccountConfig({
-    ...configOptions,
-    portalId: accountIdentifier,
-  });
+export function getConfigAccountIfExists(
```
brandenrodgers, line 172: I was reviewing one of your PR's in the CLI, and this util tripped me up a little. This is effectively the same functionality as `getConfigAccountByInferredIdentifier`. But the naming of this util doesn't make it clear that we're inferring the identifier.`

[Also here's a relevant comment from a CLI PR](https://github.com/HubSpot/hubspot-cli/pull/1392/files#r2002008951)

config/index.ts
camden11: Good call, I can add something to the readme for these. On the CLI side, we'll be setting these automatically when users use `--use-env` or `--config` paths so nothing should change there

config/utils.ts
```diff
@@ -0,0 +1,466 @@
...
+}
+
+export function getLocalConfigDefaultFilePath(): string {
+  return `${getCwd()}/${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME}`;
+}
+
+export function getConfigPathEnvironmentVariables(): {
+  useEnvironmentConfig: boolean;
+  configFilePathFromEnvironment: string | undefined;
+} {
```
joe-yeager: Could we specify a return type here

PR #232, opened 2025-02-04
feat: add fixtures to fetch scope authorization data per portal and user

b-ash:
## Description and Context

Adds various models so we can fetch scope authorization data for use in the CLI when debugging potential local dev issues.

Subsequent PR incoming in [HubSpot/hubspot-cli](https://github.com/HubSpot/hubspot-cli/) and will link accordingly.

### DIFF
--- a/api/localDevAuth.ts
+++ b/api/localDevAuth.ts
@@ -6,7 +6,6 @@ import {
   ScopeData,
   AccessTokenResponse,
   EnabledFeaturesResponse,
-  ScopeGroupAuthorization,
   ScopeAuthorizationResponse,
 } from '../types/Accounts';
 import axios from 'axios';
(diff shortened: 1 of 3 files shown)


api/localDevAuth.ts
```diff
@@ -6,6 +6,8 @@ import {
   ScopeData,
   AccessTokenResponse,
   EnabledFeaturesResponse,
+  ScopeGroupAuthorization,
```
camden11, line 9: nit: looks like this import isn't being used

camden11 (approved):
One minor comment, otherwise LGTM!

api/localDevAuth.ts
b-ash: ty! good catch

lib/personalAccessKey.ts
```diff
@@ -168,6 +168,12 @@ export async function scopesOnAccessToken(
   return (await getNewAccessTokenByAccountId(accountId)).scopeGroups;
 }
 
+export async function authorizedScopesForPortalAndUser(
```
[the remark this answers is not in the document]
b-ash: Agreed - I'll put a pin in that just for speed and we can probably refactor it later

PR #236, opened 2025-02-05
POC: Bail upload if zip is too large

joe-yeager:
## Description and Context
Add a field `projectFileZipSizeInBytes`

### DIFF
--- a/api/projects.ts
+++ b/api/projects.ts
@@ -63,6 +63,7 @@ export async function uploadProject(
   if (intermediateRepresentation) {
     const formData = {
       projectFilesZip: fs.createReadStream(projectFile),
+      projectFileZipSizeInBytes: fs.statSync(projectFile).size,
       platformVersion,
       uploadRequest: JSON.stringify({
         ...intermediateRepresentation,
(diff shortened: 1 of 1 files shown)


banderson (approved):
Much simpler than I expected! :nice:

banderson:
(also, I know this was POC/experimental, so no expectation on my end)

api/projects.ts
```diff
@@ -60,6 +60,9 @@ export async function uploadProject(
   platformVersion?: string,
   intermediateRepresentation?: unknown
 ): HubSpotPromise<UploadProjectResponse | UploadIRResponse> {
+  if (fs.statSync(projectFile).size > 100 * 1024 * 1024) {
+    throw new Error('File size is too large. The maximum file size is 100MB.');
+  }
```
joe-yeager, line 65: I know we would need to make a data driven decision on this and not just pick this value arbitrarily.  I think this would be a good method to make users aware of the size of their project.  This could also just be a warning log letting them know that they are exceeding the threshold and it may cause problems

PR #237, opened 2025-02-06
chore: Ignore dist directory

joe-yeager:
## Description and Context

Ignore `dist` directory

### DIFF
--- a/lib/ignoreRules.ts
+++ b/lib/ignoreRules.ts
@@ -8,6 +8,7 @@ const ignoreList = [
   'hubspot.config.yml',
   'hubspot.config.yaml',
   'node_modules', // dependencies
+  'dist', // Build assets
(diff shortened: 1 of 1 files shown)


joe-yeager:
> This might not be a safe change. Some devs (especially CMS devs) leverage a build step before upload. Our own [Webpack CMS CLI plugin is an example](https://www.npmjs.com/package/@hubspot/webpack-cms-plugins). But in general it's pretty common for the output of that to go into `dist`, which means this change could have some unexpected outcomes for users.
> 
> That being said, I do agree that we should be ignoring the build output directory that we use during our own local dev flows. What if we customize the output directory for Vite during local dev and then ignore that directory instead? That way there's no chance of this disrupting any existing workflows.

@brandenrodgers - That all sounds good, I opened this more or less to start the discussion

PR #242, opened 2025-02-20
fix: An argument type in the custom object API functions (part of TS conversion work in CLI)

kemmerle:
## Description and Context
As part of my PR (forthcoming) to convert the `custom-object` commands to TypeScript, I needed to change an argument type in one of the custom object API functions. 

## TODO

- [ ] Address feedback 

## Who to Notify


api/customObjects.ts
```diff
@@ -12,7 +12,7 @@ const SCHEMA_API_PATH = 'crm-object-schemas/v3/schemas';
 export function batchCreateObjects(
   accountId: number,
   objectTypeId: string,
-  objects: JSON
+  objects: Schema
```
brandenrodgers, line 15: This endpoint looks like it takes a list of object definitions. The [api catalogue](https://developers.hubspot.com/docs/reference/api/crm/objects/objects#post-%2Fcrm%2Fv3%2Fobjects%2F%7Bobjecttype%7D%2Fbatch%2Fcreate) shows an example of what it looks like. I also believe the shape for the objects will look a little different than the shape of the schemas.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "3.3.1",
+  "version": "0.1.5-experimental.0",
```
joe-yeager, line 3: Reminder to change this back before merging

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/local-dev-lib",
-  "version": "3.3.1",
+  "version": "0.1.7-experimental.0",
```
camden11, line 3: Remember to switch this back!

types/Schemas.ts
```diff
@@ -40,3 +40,25 @@ export type CreateObjectsResponse = {
...
+  primaryDisplayProperty?: string;
+  properties: [
+    {
+      isPrimaryDisplayLabel: true;
+      label: string;
+      name: string;
+    },
+  ];
+  requiredProperties: Array<string>;
+};
```
camden11, line 64: Non-blocking: it would probably make sense to keep all custom object related types in one file. Ideally that would be `CustomObjects.ts`, but that would be a breaking change. We should either put everything in `Schemas.ts` or make a note to fix this during the next major release

camden11 (approved):
LGTM but remember to update the version before you merge!

PR #247, opened 2025-03-19
Support `hs config migrate` command

kemmerle:
## Description and Context

This PR adds util functions for the new `hs config migrate` command. 

## Screenshots
See https://github.com/HubSpot/hubspot-cli/pull/1398

## Who to Notify

### DIFF
--- a/config/index.ts
+++ b/config/index.ts
@@ -98,10 +98,13 @@ export function getConfigPath(
   return config_DEPRECATED.getConfigPath(path);
 }
 
-export function configFileExists(useHiddenConfig?: boolean) {
+export function configFileExists(
+  useHiddenConfig?: boolean,
+  path?: string
+): boolean {
   return useHiddenConfig
     ? newConfigFileExists()
-    : Boolean(config_DEPRECATED.getConfigPath());
+    : Boolean(config_DEPRECATED.getConfigPath(path));
 }
 
 export function getAccountConfig(accountId?: number): CLIAccount | null {
--- a/config/index.ts
+++ b/config/index.ts
@@ -98,13 +98,10 @@ export function getConfigPath(
   return config_DEPRECATED.getConfigPath(path);
 }
 
-export function configFileExists(
-  useHiddenConfig?: boolean,
-  path?: string
-): boolean {
+export function configFileExists(useHiddenConfig?: boolean): boolean {
   return useHiddenConfig
     ? newConfigFileExists()
-    : Boolean(config_DEPRECATED.getConfigPath(path));
+    : Boolean(config_DEPRECATED.getConfigPath());
 }
 
 export function getAccountConfig(accountId?: number): CLIAccount | null {
--- a/config/migrate.ts
+++ b/config/migrate.ts
@@ -24,21 +24,24 @@ import { logger } from '../lib/logger';
 
 const i18nKey = 'config.migrate';
 
-export function getConfig(useHiddenConfig = false): Partial<CLIConfig> | null {
+export function getConfig(
+  useHiddenConfig?: boolean,
+  configPath?: string
+): Partial<CLIConfig> | null {
   if (useHiddenConfig) {
     return CLIConfiguration.config;
   }
-  return config_DEPRECATED.getAndLoadConfigIfNeeded();
+  return config_DEPRECATED.loadConfig(configPath);
 }
 
 export function getConfigPath(
-  path?: string,
+  configPath?: string,
   useHiddenConfig = false
 ): string | null {
   if (useHiddenConfig) {
     return getConfigFilePath();
   }
-  return config_DEPRECATED.getConfigPath(path);
+  return config_DEPRECATED.getConfigPath(configPath);
 }
 
 export function configFileExists(
--- a/config/migrate.ts
+++ b/config/migrate.ts
@@ -28,7 +28,6 @@ import {
   DEFAULT_PORTAL,
 } from '../constants/config';
 import { i18n } from '../utils/lang';
-import { logger } from '../lib/logger';
 
 const i18nKey = 'config.migrate';
 
@@ -78,7 +77,6 @@ function writeGlobalConfigFile(updatedConfig: CLIConfig_NEW): void {
     deleteConfigFile();
   } catch (error) {
     deleteEmptyConfigFile();
-
     throw new Error(
       i18n(`${i18nKey}.errors.writeConfig`, { configPath: GLOBAL_CONFIG_PATH }),
       { cause: error }
@@ -106,21 +104,27 @@ export function migrateConfig(
   writeGlobalConfigFile(updatedConfig);
 }
 
-function mergeConfigPropertes(
+type ConflictValue = boolean | string | number | CmsPublishMode | Environment;
+export type ConflictProperty = {
+  property: keyof CLIConfig_NEW;
+  oldValue: ConflictValue;
+  newValue: ConflictValue;
+};
+
+export function mergeConfigProperties(
   globalConfig: CLIConfig_NEW,
   deprecatedConfig: CLIConfig_DEPRECATED
-): CLIConfig_NEW {
+): {
+  initialConfig: CLIConfig_NEW;
+  conflicts: Array<ConflictProperty>;
+} {
   const propertiesToCheck: Array<keyof Partial<CLIConfig>> = [
     DEFAULT_CMS_PUBLISH_MODE,
     HTTP_TIMEOUT,
     ENV,
     HTTP_USE_LOCALHOST,
   ];
-  const conflicts: Array<{
-    property: keyof CLIConfig_NEW;
-    oldValue: boolean | string | number | CmsPublishMode | Environment;
-    newValue: boolean | string | number | CmsPublishMode | Environment;
-  }> = [];
+  const conflicts: Array<ConflictProperty> = [];
 
   propertiesToCheck.forEach(prop => {
     if (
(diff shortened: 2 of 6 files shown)


constants/config.ts
```diff
@@ -1,10 +1,13 @@
 import { i18n } from '../utils/lang';
+import os from 'os';
 
 export const DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME = 'hubspot.config.yml';
 
 export const HUBSPOT_CONFIGURATION_FOLDER = '.hscli';
 export const HUBSPOT_CONFIGURATION_FILE = 'config.yml';
 
+export const GLOBAL_CONFIG_PATH = `${os.homedir()}/${HUBSPOT_CONFIGURATION_FOLDER}/${HUBSPOT_CONFIGURATION_FILE}`;
```
joe-yeager, line 9: We should use `path.join` to make this cross platform

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+    deleteConfigFile();
+  } catch (error) {
+    deleteEmptyConfigFile();
+
+    throw new Error(
+      i18n(`${i18nKey}.errors.writeConfig`, { configPath: GLOBAL_CONFIG_PATH }),
+      { cause: error }
+    );
+  }
+}
```
joe-yeager, line 79: Can we update the existing methods instead of splitting and making new methods?

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+      if (globalConfig[prop] !== deprecatedConfig[prop]) {
+        conflicts.push({
+          property: prop,
+          oldValue: deprecatedConfig[prop]!,
+          newValue: globalConfig[prop]!,
+        });
+      }
+    } else {
+      // @ts-expect-error TODO
+      globalConfig[prop] = deprecatedConfig[prop];
```
joe-yeager, line 128: Is this resolved?

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+  ];
+  const conflicts: Array<{
+    property: keyof CLIConfig_NEW;
+    oldValue: boolean | string | number | CmsPublishMode | Environment;
+    newValue: boolean | string | number | CmsPublishMode | Environment;
+  }> = [];
+
+  propertiesToCheck.forEach(prop => {
+    if (prop in globalConfig && prop in deprecatedConfig) {
+      if (globalConfig[prop] !== deprecatedConfig[prop]) {
```
joe-yeager, line 119: Nit: these can be merged

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+
+function mergeConfigPropertes(
+  globalConfig: CLIConfig_NEW,
+  deprecatedConfig: CLIConfig_DEPRECATED
+): CLIConfig_NEW {
+  const propertiesToCheck: Array<keyof Partial<CLIConfig>> = [
+    'defaultCmsPublishMode',
+    'httpTimeout',
+    'allowUsageTracking',
+    'env',
```
joe-yeager, line 109: Can we create constants for these config values?

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+        });
+      }
+    } else {
+      // @ts-expect-error TODO
+      globalConfig[prop] = deprecatedConfig[prop];
+    }
+  });
+
+  if ('defaultAccount' in globalConfig && 'defaultPortal' in deprecatedConfig) {
+    if (globalConfig.defaultAccount !== deprecatedConfig.defaultPortal) {
```
joe-yeager, line 133: Nit: can we use constants for `defaultAccount` and `defaultPortal`. also these if statements can be merged

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (conflicts.length > 0) {
+    logger.log(
+      `The following properties have different values in the deprecated and global config files:\n${conflicts
+        .map(
+          ({ property, oldValue, newValue }) =>
+            `${property}: ${oldValue} (deprecated) vs ${newValue} (global)`
+        )
+        .join('\n')}`
+    );
+    return globalConfig;
```
joe-yeager, line 153: Nit: This return statement is redundant because we have another return as the next statement below it.

config/migrate.ts
```diff
@@ -0,0 +1,191 @@
...
+        newValue: globalConfig.defaultAccount!,
+      });
+    }
+  } else if ('defaultPortal' in deprecatedConfig) {
+    globalConfig.defaultAccount = deprecatedConfig.defaultPortal;
+  }
+
+  if (conflicts.length > 0) {
+    logger.log(
+      `The following properties have different values in the deprecated and global config files:\n${conflicts
```
joe-yeager, line 146: Should we move this to the lang file?

config/migrate.ts
kemmerle: We discussed this after sprint planning and decided it was best to create new methods for two main reasons:

1) In order to update the existing methods, I would need to change their default behavior. Currently, if the new configuration file exists (ie. `CLIConfiguration.isActive()`), we use the new config methods. If not, then we default to the `config_DEPRECATED` methods. But with the migration command, I need to be able to call the deprecated methods *even if* `CLIConfiguration.isActive()` returns true. I could edit the existing methods, but it seemed weird to do that given that our goal is to deprecate the old config. 

Another alternative for this predicament would be to directly import the `config_DEPRECATED` methods into the CLI, but because we've never exposed them before, we wanted to avoid that. 

Creating new methods seemed like the best and cleanest alternative.

2) Putting all the `migrate` methods in one file also allows us to delete them easier. `hs config migrate` by its nature is transitory--eventually, we will sunset the command (fingers crossed). When we go through and clean up the files, having every method in one file will make it easier to deprecate, rather than making changes throughout the original `config/index` file. 

Hope this helps! I can also see whether we have an AI companion writeup for that meeting! Going to resolve for now, but feel free to reopen with any more questions.

config/migrate.ts
kemmerle: I've added a note, but unfortunately, I don't believe we can resolve this without doing something gross. I've added a note to say that we can't reconcile the `CLIConfig_NEW` and `CLIConfig_DEPRECATED` types. If you have any ideas, I'm all ears 👂

joe-yeager (approved):

constants/config.ts
```diff
@@ -1,16 +1,31 @@
...
 export const DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID =
   'DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID';
 export const DEFAULT_ACCOUNT_OVERRIDE_ERROR_ACCOUNT_NOT_FOUND =
   'DEFAULT_ACCOUNT_OVERRIDE_ERROR_ACCOUNT_NOT_FOUND';
 
+export const DEFAULT_CMS_PUBLISH_MODE = 'defaultCmsPublishMode';
+export const HTTP_TIMEOUT = 'httpTimeout';
+export const ENV = 'env';
+export const HTTP_USE_LOCALHOST = 'httpUseLocalhost';
+export const DEFAULT_ACCOUNT = 'defaultAccount';
```
brandenrodgers, line 26: I wonder if we want to make the decision for the user in the case where there's a conflict between `defaultAccount` and `defaultPortal`. We never want `defaultPortal` in the global config because it's deprecated.

I'm realizing a similar thing applies to `accounts` vs. `portals`. Let me know if this migration logic is already handling it though

config/migrate.ts
```diff
@@ -0,0 +1,209 @@
...
+      }));
+
+    if (newAccounts.length > 0) {
+      globalConfig.accounts.push(...newAccounts);
+    }
+  }
+
+  return {
+    finalConfig: globalConfig,
+    skippedAccountIds: deprecatedConfig.portals
```
brandenrodgers, line 192: Not a blocker, but you could DRY this up a little by generating the `skippedAccounIds` array in the `filter()`

config/migrate.ts
```diff
@@ -0,0 +1,209 @@
...
+export function deleteConfigFile(useHiddenConfig = false): void {
+  if (useHiddenConfig) {
+    newDeleteConfigFile();
+  }
+  config_DEPRECATED.deleteConfigFile();
+}
+
+function writeGlobalConfigFile(updatedConfig: CLIConfig_NEW): void {
+  const updatedConfigJson = JSON.stringify(updatedConfig);
+  createEmptyConfigFile({}, true);
```
brandenrodgers, line 72: Do we need to create a new empty config file, even if one already exists?

config/migrate.ts
```diff
@@ -0,0 +1,209 @@
...
+  useHiddenConfig?: boolean,
+  configPath?: string
+): boolean {
+  return useHiddenConfig
+    ? newConfigFileExists()
+    : Boolean(config_DEPRECATED.getConfigPath(configPath));
+}
+
+export function deleteConfigFile(useHiddenConfig = false): void {
+  if (useHiddenConfig) {
```
brandenrodgers, line 64: Will we ever want this to delete the global config file? I'm wondering if we can get rid of this util and just call `config_DEPRECATED.deleteConfigFile()` directly wherever we need it

config/migrate.ts
```diff
@@ -0,0 +1,209 @@
...
+  configPath?: string,
+  useHiddenConfig = false
+): string | null {
+  if (useHiddenConfig) {
+    return getConfigFilePath();
+  }
+  return config_DEPRECATED.getConfigPath(configPath);
+}
+
+export function configFileExists(
```
brandenrodgers, line 54: If you want, we could probably just update the existing `configFileExists` in config/index to support the `configPath` field. I'm guessing you did it here to avoid tampering with anything in that file, but this change seems fairly safe to me since it would be backwards compatible

config/migrate.ts
```diff
@@ -0,0 +1,209 @@
...
+    DEFAULT_CMS_PUBLISH_MODE,
+    HTTP_TIMEOUT,
+    ENV,
+    HTTP_USE_LOCALHOST,
+  ];
+  const conflicts: Array<ConflictProperty> = [];
+
+  propertiesToCheck.forEach(prop => {
+    if (prop in globalConfig && prop in deprecatedConfig) {
+      if (force) {
```
brandenrodgers, line 132: Not a blocker, you could also DRY this up with `if (force || globalConfig[prop] == deprecatedConfig[prop])`

lang/en.json
```diff
@@ -311,6 +311,13 @@
         "missingEnv": "Unable to load config from environment variables: Missing env",
         "unknownAuthType": "Unable to load config from environment variables: Unknown auth type"
       }
+    },
+    "migrate": {
+      "conflicts": "The following properties have different values in the deprecated and global config files: \n{{ formattedConflicts }}",
```
brandenrodgers, line 316: Looks like this copy in `conflicts` isn't needed in the latest iteration

constants/config.ts
kemmerle: The migration logic handles both, so the properties `defaultPortal` and `portals` will never appear in the new config. 

We do prompt the user if the values of `defaultAccount` and `defaultPortal` are different, and the user has the option to write over the value of `defaultAccount` but not the property name. Let me know if that clears things up, happy to talk it through in a huddle.

config/index.ts
```diff
@@ -98,10 +98,13 @@ export function getConfigPath(
...
 
-export function configFileExists(useHiddenConfig?: boolean) {
+export function configFileExists(
+  useHiddenConfig?: boolean,
+  path?: string
+): boolean {
   return useHiddenConfig
     ? newConfigFileExists()
-    : Boolean(config_DEPRECATED.getConfigPath());
+    : Boolean(config_DEPRECATED.getConfigPath(path));
```
camden11, line 107: I'm not sure if this will actually check if the config file exists at the given path - it looks like `config_DEPRECATED` just returns the path if one is passed in

config/migrate.ts
```diff
@@ -0,0 +1,197 @@
...
+const i18nKey = 'config.migrate';
+
+export function getConfig(
+  useHiddenConfig?: boolean,
+  configPath?: string
+): Partial<CLIConfig> | null {
+  if (useHiddenConfig) {
+    return CLIConfiguration.config;
+  }
+  return config_DEPRECATED.loadConfig(configPath);
```
camden11, line 37: Is this calling `loadConfig` because it may not have been loaded if the user is using a hidden config?

kemmerle: Yes, exactly. Confusingly, the deprecated `loadConfig` file both loads and gets the deprecated config, so it should really be named something like `loadAndGetConfig`. But I didn't see the point in changing the name now, given that we're deprecating it.

PR #248, opened 2025-03-21
feat: Add http methods required for migrating to new platform version

joe-yeager:
## Description and Context

Updates the methods used for migration to use the new V2 endpoints for the new platform versions, and add new endpoints as needed.


api/projects.ts
```diff
@@ -351,10 +403,21 @@ export function migrateApp(
...
 ): HubSpotPromise<PollAppResponse> {
+  // TODO: This is a WIP, we need to wait for the BE team to implement it for the new platform version
+  if (
+    targetPlatformVersion === PLATFORM_VERSIONS.unstable ||
+    targetPlatformVersion === PLATFORM_VERSIONS.v2025_2
+  ) {
+    return http.get<PollAppResponse>(accountId, {
+      url: `${MIGRATIONS_API_PATH_V2}/migrations/${id}`,
+    });
+  }
```
joe-yeager, line 417: Waiting on the BE team to implement this endpoint and then I will update the URL and the Response shape

brandenrodgers (approved):

PR #249, opened 2025-03-25
Support `hs account` bucket refactor

kemmerle:
## Description and Context

This PR adds functions that support refactoring the `hs account` bucket to account for `.hsaccount` override files with the new config. 

## Screenshots
See https://github.com/HubSpot/hubspot-cli/pull/1407

## TODO

- [ ] Address feedback

## Who to Notify

### DIFF
--- a/config/index.ts
+++ b/config/index.ts
@@ -244,9 +244,8 @@ export function getConfigDefaultAccount(): string | number | null | undefined {
 export function getDisplayDefaultAccount(): string | number | null | undefined {
   if (CLIConfiguration.isActive()) {
     return CLIConfiguration.config?.defaultAccount;
-  } else {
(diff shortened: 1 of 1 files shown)


config/index.ts
```diff
@@ -241,6 +241,18 @@ export function getConfigDefaultAccount(): string | number | null | undefined {
   return config_DEPRECATED.getConfigDefaultAccount();
 }
 
+export function getDisplayDefaultAccount(
+  isOverride: boolean
```
camden11, line 245: Could you simplify this by removing the `isOverride` option and just calling `getConfigDefaultAccount` when you need the overridden account?

camden11 (approved):

PR #252, opened 2025-04-02
Add timeout to HubSpotHttpError

camden11:
## Description and Context
This adds the `timeout` field from `AxiosError` to `HubSpotHttpError`, so that the CLI can check it and determine if timeouts were caused by the config default timeout.

## Who to Notify


kemmerle (approved):
LGTM!

joe-yeager (approved):

PR #254, opened 2025-04-04
Dynamic config migrations

camden11:
## Description and Context
In an attempt to not let the config revamp work get too stale, I've updated the `dynamic-config` branch to apply the new config logic to @kemmerle's migration work. The fact that the global and deprecated config are both converted to the same object shape makes merging them much simpler. Like the rest of the config revamp work, this will be breaking for the CLI, but we'll cross that bridge when we get there

Note: merging this into my `dynamic-config` branch, not main

## Who to Notify

@kemmerle @brandenrodgers @joe-yeager

### DIFF
--- a/config/migrate.ts
+++ b/config/migrate.ts
@@ -18,7 +18,7 @@ import {
 } from './utils';
 import { ValueOf } from '../types/Utils';
 
-function getConfigAtPath(path: string): HubSpotConfig {
+export function getConfigAtPath(path: string): HubSpotConfig {
   const configFileSource = readConfigFile(path);
 
   return parseConfig(configFileSource);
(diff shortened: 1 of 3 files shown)


config/migrate.ts
```diff
@@ -1,218 +1,119 @@
...
+      DEFAULT_CMS_PUBLISH_MODE,
+      HTTP_TIMEOUT,
+      ENV,
+      HTTP_USE_LOCALHOST,
+      ALLOW_USAGE_TRACKING,
+      DEFAULT_ACCOUNT,
+    ] as const;
+
+    propertiesToCheck.forEach(prop => {
+      if (toConfig[prop] !== fromConfig[prop]) {
```
brandenrodgers, line 68: The from config value could be undefined here (I think). Should we only count it as a conflict if it has a value?

camden11: Oh yes good call

brandenrodgers (approved):
This was the first time I've seen the `||=` operator. That's pretty neat

PR #255, opened 2025-04-08
chore: Update to async endpoints

joe-yeager:
## Description and Context

Updates the types as needed to account for the new APIs

### DIFF
--- a/api/projects.ts
+++ b/api/projects.ts
@@ -9,9 +9,6 @@ import {
   FetchPlatformVersionResponse,
   WarnLogsResponse,
   UploadIRResponse,
-  ListAppsResponse,
-  InitializeMigrationResponse,
-  ContinueMigrationResponse,
 } from '../types/Project';
 import { Build, FetchProjectBuildsResponse } from '../types/Build';
 import {
@@ -23,9 +20,7 @@ import {
   MigrateAppResponse,
   CloneAppResponse,
   PollAppResponse,
-  MigrationStatus,
 } from '../types/Migration';
-import { PLATFORM_VERSIONS } from '../constants/projects';
 
 const PROJECTS_API_PATH = 'dfs/v1/projects';
 const DEVELOPER_FILE_SYSTEM_PATH = 'dfs/v1';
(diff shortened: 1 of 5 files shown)


types/Migration.ts
```diff
@@ -7,6 +7,8 @@ export const MIGRATION_STATUS = {
   PREPARING: 'PREPARING',
   PENDING: 'PENDING',
   SUCCESS: 'SUCCESS',
+  IN_PROGRESS: 'IN_PROGRESS',
+  INPUT_REQUIRED: 'INPUT_REQUIRED',
```
joe-yeager, line 11: I'm curious if anyone else has encountered this issue, but importing this object in the CLI breaks the `ts-node` execution that we use to run commands locally.  I.E. if you import this object in a command (say hs app migrate) and then try to run `yarn hs app migrate` you get:

```
import { ValueOf } from './Utils';
^^^^^^

SyntaxError: Cannot use import statement outside a module

```

I haven't really had the time to look into the cause, but that's a head scratched.  It appears to be the `as const` syntax

api/projects.ts
```diff
@@ -347,31 +348,36 @@ export async function listAppsForMigration(
   });
 }
 
-export async function beginMigration(
+function mapPlatformVersionToEnum(platformVersion: string): string {
```
brandenrodgers, line 351: Would we be future proofing ourselves by putting this logic somewhere else? If we limit the amount of logic that happens in these api utils, we won't have to worry about making (potentially breaking) logic changes down the road.

Or are these enums very specific to the migrations api?

types/Migration.ts
brandenrodgers: I was messing around with this in Cursor and was able to fix it with this change in the package.json exports:

```json
// Before
{
  "./types/*": "./types/*.d.ts"
}
// After
{
  "./types/*": {
    "types": "./types/*.d.ts",
    "default": "./types/*.js"
  }
}
```

brandenrodgers (approved):
Just a few suggestions but no blockers

types/Migration.ts
joe-yeager: Sweet that fixed it! That was bothering me so much haha

api/projects.ts
joe-yeager: > Or are these enums very specific to the migrations api?

Yeah, they seem to be specific to migrations ATM.  This is the only place we are using it ATM.  Since this particular method isn't exposed, we could move it in the future if we find a reason to reuse it.

PR #257, opened 2025-04-10
Fix bug when merging configs (global config work)

kemmerle:
## Description and Context
@camden11 discovered a bug when merging deprecated and global configs in the `hs config migrate` command: Properties that existed in the deprecated config but not in the global were not being properly migrated to the global config.  This PR fixes that bug. 

## TODO

- [ ] Address feedback 

## Who to Notify

### DIFF
--- a/config/migrate.ts
+++ b/config/migrate.ts
@@ -130,8 +130,12 @@ export function mergeConfigProperties(
   const conflicts: Array<ConflictProperty> = [];
 
   propertiesToCheck.forEach(prop => {
-    if (prop in globalConfig && prop in deprecatedConfig) {
-      if (force || globalConfig[prop] === deprecatedConfig[prop]) {
+    if (prop in deprecatedConfig) {
+      if (
+        force ||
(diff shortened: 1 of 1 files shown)


camden11 (approved):

config/migrate.ts
```diff
@@ -130,8 +130,12 @@ export function mergeConfigProperties(
...
 
   propertiesToCheck.forEach(prop => {
-    if (prop in globalConfig && prop in deprecatedConfig) {
-      if (force || globalConfig[prop] === deprecatedConfig[prop]) {
+    if (prop in deprecatedConfig) {
+      if (
+        force ||
+        !(prop in globalConfig) ||
+        globalConfig[prop] === deprecatedConfig[prop]
+      ) {
```
joe-yeager, line 138: Could be some tests to cover this function and make sure we don't add regressions?

kemmerle: I'll add tests as a fast follow--I want to get this out today before tomorrow's launch of global config.

PR #260, opened 2025-04-16
feat: Add the ability to debug log all the network requests

joe-yeager:
## Description and Context

Add the ability to debug log all of the network requests


brandenrodgers (approved):
Seems like a cool idea! I'm down to give it a try

PR #261, opened 2025-04-22
Fix the return type of the create dev test account api

brandenrodgers:
## Description and Context

The return type isn't the same as the dev test account shape. It includes a personal access key that we can immediately store in the CLI config. This prevents us from having to prompt the user for the key after creating the account.


kemmerle (approved):
Makes sense! :shipit:

joe-yeager (approved):

PR #263, opened 2025-04-25
Merge networking logging to right branch

joe-yeager:
## Description and Context

I had a PR for this previously and the target branch didn't update, so re opening targeting main


joe-yeager:
This was previously approved over here: https://github.com/HubSpot/hubspot-local-dev-lib/pull/260

I just had the wrong branch targeted

PR #265, opened 2025-05-06
fix: Check if string is defined before attempting to access string methods

joe-yeager:
## Description and Context
Check if the string is defined before calling `str.replace`

### DIFF
--- a/lib/__tests__/text.test.ts
+++ b/lib/__tests__/text.test.ts
@@ -1,4 +1,4 @@
-import { commaSeparatedValues } from '../text';
+import { commaSeparatedValues, toKebabCase } from '../text';
 
 describe('lib/text', () => {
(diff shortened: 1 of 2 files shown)


lib/__tests__/text.test.ts
```diff
@@ -19,5 +19,48 @@ describe('lib/text', () => {
...
+      const res = commaSeparatedValues(['only']);
+
+      expect(res).toBe('only');
+    });
+
+    it('handles a two item array', () => {
+      const res = commaSeparatedValues(['first', 'second']);
+
+      expect(res).toBe('first and second');
+    });
```
joe-yeager, line 33: `claude` found these gaps in testing 🙌

camden11 (approved):

PR #266, opened 2025-05-07
fix: Check if errors is an array before attempting to iterate it

joe-yeager:
## Description and Context
Fixes an issue when the upload endpoint sometimes returns a `string` instead an array of `ErrorDetails`


brandenrodgers (approved):
👌

PR #270, opened 2025-05-12
Fix: Account names cannot be numbers

kemmerle:
## Description and Context
The DevRel team discovered a bug, where they could not name accounts numbers, (ie. `2024`, `57`, etc). This was because we were using a regular expression to check with the `nameOrId` was a number--if it was, we then converted it to a number and treated it as the account ID. Now that we've implemented TypeScript in LDL and the CLI, we no longer need to perform this kind of check. We always know that the account name will be a string and the account ID is a number. 

## Screenshots
Before:

Screenshot 2025-05-12 at 10 34 21 AM

Screenshot 2025-05-12 at 10 33 28 AM

After:

Screenshot 2025-05-12 at 10 49 22 AM

Screenshot 2025-05-12 at 10 50 49 AM

## TODO

- [x] Address feedback

## Who to Notify

### DIFF
--- a/config/CLIConfiguration.ts
+++ b/config/CLIConfiguration.ts
@@ -204,12 +204,10 @@ class _CLIConfiguration {
       return null;
     }
 
-    if (typeof nameOrIdToCheck === 'number') {
-      accountId = nameOrIdToCheck;
-    } else if (/^\d+$/.test(nameOrIdToCheck)) {
-      accountId = parseInt(nameOrIdToCheck, 10);
-    } else {
+    if (typeof nameOrIdToCheck === 'string') {
       name = nameOrIdToCheck;
+    } else if (typeof nameOrIdToCheck === 'number') {
+      accountId = nameOrIdToCheck;
     }
 
     if (name) {
--- a/config/CLIConfiguration.ts
+++ b/config/CLIConfiguration.ts
@@ -206,12 +206,19 @@ class _CLIConfiguration {
 
     if (typeof nameOrIdToCheck === 'string') {
       name = nameOrIdToCheck;
+      if (/^\d+$/.test(nameOrIdToCheck)) {
+        accountId = parseInt(nameOrIdToCheck, 10);
+      }
     } else if (typeof nameOrIdToCheck === 'number') {
       accountId = nameOrIdToCheck;
     }
 
     if (name) {
-      return this.config.accounts.find(a => a.name === name) || null;
+      return (
+        this.config.accounts.find(a => a.name === name) ||
+        this.config.accounts.find(a => accountId === a.accountId) ||
+        null
+      );
     } else if (accountId) {
       return this.config.accounts.find(a => accountId === a.accountId) || null;
     }
--- a/config/CLIConfiguration.ts
+++ b/config/CLIConfiguration.ts
@@ -213,17 +213,17 @@ class _CLIConfiguration {
       accountId = nameOrIdToCheck;
     }
 
+    let account = null;
     if (name) {
-      return (
-        this.config.accounts.find(a => a.name === name) ||
-        this.config.accounts.find(a => accountId === a.accountId) ||
-        null
-      );
-    } else if (accountId) {
-      return this.config.accounts.find(a => accountId === a.accountId) || null;
+      account = this.config.accounts.find(a => a.name === name) || null;
     }
 
-    return null;
+    if (accountId && !account) {
+      account =
+        this.config.accounts.find(a => accountId === a.accountId) || null;
+    }
+
+    return account;
   }
 
   isConfigFlagEnabled(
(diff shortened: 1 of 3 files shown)


config/config_DEPRECATED.ts
```diff
@@ -408,12 +408,10 @@ export function getAccountId(nameOrId?: string | number): number | undefined {
   function setNameOrAccountFromSuppliedValue(
     suppliedValue: string | number
   ): void {
-    if (typeof suppliedValue === 'number') {
-      accountId = suppliedValue;
-    } else if (/^\d+$/.test(suppliedValue)) {
-      accountId = parseInt(suppliedValue, 10);
-    } else {
+    if (typeof suppliedValue === 'string') {
       name = suppliedValue;
```
brandenrodgers, line 412: I think in some cases, our `--account` flag supports either a string (`my-prod-account`) or a number that represents the account Id (`1248533`).  I think in both cases it will show up in this util as a string. I'm wondering with these changes would I still be able to do something like `hs account use 123456` where `123456` is my account id?

For example, here's the output of `hs account use --help`
image

config/CLIConfiguration.ts
```diff
@@ -204,16 +204,21 @@ class _CLIConfiguration {
...
-    if (typeof nameOrIdToCheck === 'number') {
-      accountId = nameOrIdToCheck;
-    } else if (/^\d+$/.test(nameOrIdToCheck)) {
-      accountId = parseInt(nameOrIdToCheck, 10);
-    } else {
+    if (typeof nameOrIdToCheck === 'string') {
       name = nameOrIdToCheck;
+      if (/^\d+$/.test(nameOrIdToCheck)) {
+        accountId = parseInt(nameOrIdToCheck, 10);
+      }
```
kemmerle, line 211: If the 'nameOrIdToCheck` is a string **and** numeric, we set both `name` and `accountId` to that value. 

Then, we check whether it is present in `this.config.accounts` using both the `name` and `accountId` property. 

This allows us to account for two scenarios:
1) A customer names an account something numeric (ie. "2024", "1956" etc).
2) A customer supplies an account ID to a command (`hs account use 12345678`), which the CLI interprets it as a string.

config/CLIConfiguration.ts
```diff
@@ -206,12 +206,19 @@ class _CLIConfiguration {
...
     } else if (typeof nameOrIdToCheck === 'number') {
       accountId = nameOrIdToCheck;
     }
 
     if (name) {
-      return this.config.accounts.find(a => a.name === name) || null;
+      return (
```
brandenrodgers, line 217: This totally works, but I'm curious why you went with different logic here than over in the deprecated config file. Could we pick one and use the pattern in both places?

kemmerle: The logic is the same, just looks a bit different. If there's a name, we use it and prefer it, falling back to the version that's parsed into an integer if that's possible.

The reason they look different is because in the deprecated case portalId can be undefined. While in the global config case, the accountId cannot be undefined. So we need an extra check in the portalId case which would be redundant in the accountId case.

The deprecated code also needs to get the portalId before returning, while the global config returns the entire account.

If the deprecated code follows the pattern of the global config code it will match undefined portalIds. We can contort the global config code to look more like the old one, but they will still be different and we're just making the code that we're going to keep for the long term more complicated going forward.

I factored out the accounts check in the deprecated code, that does make them a little bit more similar.

brandenrodgers: That makes sense, I'm mostly wondering why the logic is structured differently between the two. In this new config file we're duplicating the logic that looks for accounts via account ID. We have logic in this IF case that's duplicated in the ELSE case, and it seems like the pattern used in the deprecated config file would fix it.

Is there any reason this couldn't be something along the line of this:
```js
let account = null:

if (name) {
  account = this.config.accounts.find(a => a.name === name) || null;
}

if (accountId && !account) {
  account = this.config.accounts.find(a => accountId === a.accountId) || null;
}

return account;
``` 

This matches the other pattern for consistency and DRY's things up.

config/CLIConfiguration.ts
```diff
@@ -204,21 +204,26 @@ class _CLIConfiguration {
...
+    if (typeof nameOrIdToCheck === 'string') {
       name = nameOrIdToCheck;
+      if (/^\d+$/.test(nameOrIdToCheck)) {
+        accountId = parseInt(nameOrIdToCheck, 10);
+      }
+    } else if (typeof nameOrIdToCheck === 'number') {
+      accountId = nameOrIdToCheck;
     }
 
+    let account = null;
```
brandenrodgers, line 216: We should probably type this just to give some type safety

brandenrodgers (approved):

PR #274, opened 2025-05-21
feat: Support extracting multiple `sourceDir` locations in `extractZipArchive`

joe-yeager:
## Description and Context

Update `extractZipArchive` method to take a array of `sourceDir` and copy all of the locations to the destination directoy.

### DIFF
--- a/lib/archive.ts
+++ b/lib/archive.ts
@@ -93,7 +93,9 @@ async function copySourceToDest(
     const sourceDirs = [];
 
     if (sourceDir) {
-      sourceDirs.push(...(Array.isArray(sourceDir) ? sourceDir : [sourceDir]));
+      sourceDirs.push(
+        ...(Array.isArray(sourceDir) ? new Set(sourceDir) : [sourceDir])
+      );
     }
 
     if (sourceDirs.length === 0) {
@@ -116,7 +118,7 @@ async function copySourceToDest(
             file => path.relative(projectSrcDir, file)
           );
 
-          // Files that exist in the same positions in both the
+          // Find files that exist in the same positions in both directories
           collisions = existingFiles.filter(currentFile =>
             newFiles.includes(currentFile)
           );
--- a/lib/archive.ts
+++ b/lib/archive.ts
@@ -88,13 +88,22 @@ async function copySourceToDest(
       srcDirPath.push(rootDir);
     }
 
+    const foo = [];
+
     if (sourceDir) {
-      srcDirPath.push(sourceDir);
+      foo.push(...(Array.isArray(sourceDir) ? sourceDir : [sourceDir]));
     }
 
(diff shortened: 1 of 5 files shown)


types/Github.ts
```diff
@@ -77,6 +77,6 @@ export type CloneGithubRepoOptions = {
   type?: string; // The type of asset being downloaded. Used for logging
   branch?: string; // Repo branch
   tag?: string; // Repo tag
-  sourceDir?: string; // The directory within the downloaded repo to write after extraction
+  sourceDir?: string | string[]; // The directory within the downloaded repo to write after extraction
```
joe-yeager, line 80: I wanted to change this variable name, but that is a breaking change, and I didn't want to take on the complexity of having both `sourceDir` and `sourceDirs`

lib/archive.ts
```diff
@@ -88,13 +90,53 @@ async function copySourceToDest(
       srcDirPath.push(rootDir);
     }
 
+    const sourceDirs = [];
+
     if (sourceDir) {
-      srcDirPath.push(sourceDir);
+      sourceDirs.push(...(Array.isArray(sourceDir) ? sourceDir : [sourceDir]));
```
brandenrodgers, line 96: This might be an unnecessary check, but is there any value in making sure the sourceDirs are unique?

lib/archive.ts
```diff
@@ -88,13 +90,53 @@ async function copySourceToDest(
...
+          typeof handleCollision === 'function'
+        ) {
+          const existingFiles = (await walk(dest, ['node_modules'])).map(file =>
+            path.normalize(path.relative(dest, file))
+          );
+          const newFiles = (await walk(projectSrcDir, ['node_modules'])).map(
+            file => path.relative(projectSrcDir, file)
+          );
+
+          // Files that exist in the same positions in both the
```
brandenrodgers, line 119: Nit, but doesn't look like this comment was finished

lib/archive.ts
```diff
@@ -88,13 +90,53 @@ async function copySourceToDest(
...
+        if (
+          collisions.length &&
+          handleCollision &&
+          typeof handleCollision === 'function'
+        ) {
+          handleCollision({
+            dest,
+            src: projectSrcDir,
+            collisions,
+          });
```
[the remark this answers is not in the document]
brandenrodgers: Are the handlers always synchronous?

lib/archive.ts
joe-yeager: Good catch, I'll add a dedupe

lib/archive.ts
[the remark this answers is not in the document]
joe-yeager: I updated it to support async functions.  I swore in the past it was more complex to tell if a method is sync or async, but I guess it's valid to just `await` the function and if it's sync it will just resolve immediately.

PR #284, opened 2025-07-17
Support app install endpoints

brandenrodgers:
## Description and Context

Supporting the new app install endpoints. These support installing oauth apps into dev test accounts.
Details here: https://tools.hubteamqa.com/api-catalog/services/DeveloperTestPortalsService/v3/spec/internal

### DIFF
--- a/api/developerTestAccounts.ts
+++ b/api/developerTestAccounts.ts
@@ -9,6 +9,7 @@ import {
   DeveloperTestAccountConfig,
   CreateDeveloperTestAccountV3Response,
   InstallOauthAppIntoDeveloperTestAccountResponse,
+  TestPortalStatusResponse,
 } from '../types/developerTestAccounts';
 import { SANDBOX_TIMEOUT } from '../constants/api';
 import { Environment } from '../types/Config';
@@ -109,8 +110,8 @@ export function fetchDeveloperTestAccountOauthAppInstallStatus(
   accountId: number,
   projectName: string,
   appUId: string
-): HubSpotPromise<{ status: 'IN_PROGRESS' | 'SUCCESS' }> {
-  return http.post<{ status: 'IN_PROGRESS' | 'SUCCESS' }>(accountId, {
+): HubSpotPromise<TestPortalStatusResponse> {
+  return http.post<TestPortalStatusResponse>(accountId, {
     url: `${TEST_ACCOUNTS_API_PATH_V3}/install-apps/readiness`,
     data: {
       developerQualifiedSymbol: { developerSymbol: appUId, projectName },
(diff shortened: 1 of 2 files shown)


api/developerTestAccounts.ts
```diff
@@ -88,6 +89,36 @@ export function fetchDeveloperTestAccountData(
   return axios<DeveloperTestAccount>(reqWithToken);
 }
 
+export function installOauthAppIntoDeveloperTestAccount(
```
brandenrodgers, line 92: Open to better name suggestions, but this util is nuanced so I figured it would be best to be hyper specific here to avoid confusion

api/developerTestAccounts.ts
```diff
@@ -88,6 +89,36 @@ export function fetchDeveloperTestAccountData(
...
+    },
+    timeout: SANDBOX_TIMEOUT,
+  });
+}
+
+export function fetchDeveloperTestAccountOauthAppInstallStatus(
+  accountId: number,
+  projectName: string,
+  appUId: string
+): HubSpotPromise<{ status: 'IN_PROGRESS' | 'SUCCESS' }> {
```
joe-yeager, line 112: Nit: We should have constants for these

joe-yeager: Also, can this fail?

brandenrodgers: It actually can't ever return a fail status. The endpoint only returns in_progress or success statuses. It's something to do with the service being a wrapper around some other functionality (I don't fully understand)

But good call on the types. I'll add that 👍

api/developerTestAccounts.ts
```diff
@@ -88,11 +90,41 @@ export function fetchDeveloperTestAccountData(
   return axios<DeveloperTestAccount>(reqWithToken);
 }
 
+export function installOauthAppIntoDeveloperTestAccount(
+  accountId: number,
```
camden11, line 94: nit but calling this `parentAccountId` would be a little clearer i think

types/developerTestAccounts.ts
```diff
@@ -41,3 +41,17 @@ export type DeveloperTestAccountConfig = {
...
+  authCodes: Array<{
+    developerQualifiedSymbol: {
+      projectName: string;
+      developerSymbol: string;
+    };
+    authCode: string;
+  }>;
+};
+
+export type TestPortalStatusResponse = {
```
camden11, line 55: Saw your comment that these are the only response types, but what happens when this fails then? It seems like this is something that could fail pretty frequently since it wont work unless users backends are running

brandenrodgers: Yeah confirmed with Caitlyn that these are the only two states that the BE can ever return. The only other possible state is a 500, but that wouldn't surface as a status type

joe-yeager (approved):

PR #287, opened 2025-07-23
Support for V2 standard sandboxes

adamawang:
## Description and Context

This PR adds the plumbing for creating v2 sandboxes and all its types. Also includes a new endpoint for generating a PAK which is now different in the v2 sandbox experience. 

Previously, we had a `generatePersonalAccessKey` param in the create v1 sandbox flow but that has now changed for v2. 

This PR is linked to the following hubspot-cli-private PR: 

## Screenshots

n/a


types/developerTestAccounts.ts
```diff
@@ -22,6 +22,7 @@ export type CreateDeveloperTestAccountResponse = {
 export type CreateDeveloperTestAccountV3Response = {
   id: number;
   accountName: string;
+  personalAccessKey?: string;
```
adamawang, line 25: TS was complaining about this in the build so I fixed. Unrelated to this PR though

adamawang: hmm TS error is now gone after removal, I must have been in a weird state

types/Sandbox.ts
```diff
@@ -164,3 +193,10 @@ export type SandboxType = {
 export type FetchTypesResponse = {
   results: Array<SandboxType>;
 };
+
+export type PersonalAccessKey = {
```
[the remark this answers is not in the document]
adamawang: Yah good call, pushing a fix!

PR #291, opened 2025-07-30
Add config flag functionality

camden11:
## Description and Context
This updates both the new and deprecated config to support a new `flags` field that allows the CLI (and anything else) to set, check, and delete string flags. This will let us easily persist CLI state between runs without the need to modify local-dev-lib in the future. Currently, the field is called `flags` in the config, but referred to as `localStateFlags` otherwise. I'd prefer to just call it flags, but we already have the `isConfigFlagEnabled` function (which really deals more with config _settings_ than flags). It might make sense to rename everything accordingly in the next major release, but didn't want to add a breaking change now.

The impetus for this change was the need to track whether or not a use has _ever_ viewed Local Dev UI, and show them the welcome screen if they haven't. This felt important to track, but not quite deserving of its own field in the config. In the future we can use the flags array for similar things.

Since the flags will be supported in the deprecated config, I also had to account for how disparate flags would be handled in config migrations. There weren't any tests for config migrate functionality so I added some.

## TODO
* Maybe rename some things, either now or before the next major release

## Who to Notify
@brandenrodgers @joe-yeager @lkosarpersonal

### DIFF
--- a/config/migrate.ts
+++ b/config/migrate.ts
@@ -131,7 +131,7 @@ export function mergeConfigProperties(
   initialConfig: CLIConfig_NEW;
   conflicts: Array<ConflictProperty>;
 } {
-  const propertiesToCheck: Array<keyof Partial<CLIConfig>> = [
+  const propertiesToCheck: Array<keyof Omit<CLIConfig, 'flags'>> = [
     DEFAULT_CMS_PUBLISH_MODE,
     HTTP_TIMEOUT,
     ENV,
@@ -158,6 +158,15 @@ export function mergeConfigProperties(
       }
     }
   });
+
+  if (globalConfig.flags || deprecatedConfig.flags) {
+    globalConfig.flags = Array.from(
+      new Set([
+        ...(globalConfig.flags || []),
+        ...(deprecatedConfig.flags || []),
+      ])
+    );
+  }
 
   if (
     DEFAULT_ACCOUNT in globalConfig &&
--- a/config/CLIConfiguration.ts
+++ b/config/CLIConfiguration.ts
@@ -695,6 +695,36 @@ class _CLIConfiguration {
     }
     return config;
   }
+
+  hasStateFlag(flag: string): boolean {
+    if (!this.config) {
+      return false;
+    }
+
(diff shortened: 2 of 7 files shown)


config/migrate.ts
```diff
@@ -159,6 +159,15 @@ export function mergeConfigProperties(
     }
   });
 
+  if (globalConfig.flags || deprecatedConfig.flags) {
```
brandenrodgers, line 162: Should we handle conflicts (like, push them to the conflicts array)? Alternatively, since global config is newer maybe it would make sense to let any global config flags override the deprecated config flags?

brandenrodgers:
Just one comment about config migrations, but this lgtm 👍

config/migrate.ts
brandenrodgers: Makes sense to me. Defaulting to the new config settings seems fine

joe-yeager:
> The impetus for this change was the need to track whether or not a use has ever viewed Local Dev UI, and show them the welcome screen if they haven't. This felt important to track, but not quite deserving of its own field in the config. In the future we can use the flags array for similar things.

@camden11 - Could we set a flag in the browsers [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) when we show the welcome page the first time?

camden11:
@joe-yeager We decided yesterday to store this flag locally to not run the risk of showing the welcome screen multiple times if users switch browsers, clears storage, etc. We figured the welcome screen will do more harm than good if it keeps showing up when users don't want to see it

@brandenrodgers Made that update to migrate behavior!

joe-yeager:
> @joe-yeager We decided yesterday to store this flag locally to not run the risk of showing the welcome screen multiple times if users switch browsers, clears storage, etc. We figured the welcome screen will do more harm than good if it keeps showing up when users don't want to see it

Won't users that have multiple configs see the welcome screen every time they switch config files? Maybe we should leverage both for a complete solution.

camden11:
That's a good point, we were thinking in terms of global config. I do think that this is a better long term solution but we can also add a backup check in the browser for users on the deprecated config

PR #292, opened 2025-07-31
Add autoOpenLocalDevSessionInBrowser option to config

camden11:
## Description and Context
This adds a new config setting/flag `autoOpenLocalDevSessionInBrowser`. This will be used by the CLI to allow users to prevent Local Dev UI from opening automatically when they run `hs project dev`

## Who to Notify
@brandenrodgers @joe-yeager @lkosarpersonal

### DIFF
--- a/types/Config.ts
+++ b/types/Config.ts
@@ -13,6 +13,7 @@ export interface CLIConfig_NEW {
   httpTimeout?: number;
   env?: Environment;
   httpUseLocalhost?: boolean;
+  autoOpenLocalDevSessionInBrowser?: boolean;
 }
 
 export interface CLIConfig_DEPRECATED {
@@ -25,6 +26,7 @@ export interface CLIConfig_DEPRECATED {
   httpTimeout?: number;
   env?: Environment;
   httpUseLocalhost?: boolean;
+  autoOpenLocalDevSessionInBrowser?: boolean;
 }
 
 export type CLIConfig = CLIConfig_NEW | CLIConfig_DEPRECATED;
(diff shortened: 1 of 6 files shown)


types/Config.ts
```diff
@@ -13,6 +13,7 @@ export interface CLIConfig_NEW {
   httpTimeout?: number;
   env?: Environment;
   httpUseLocalhost?: boolean;
+  autoOpenLocalDevSessionInBrowser?: boolean;
```
joe-yeager, line 16: What do you think about making this a generic option?  Something along the lines of `autoOpenBrowser`?  Then we could make an abstraction over the `open` library that is basically:

```typescript
// Horrible function name, but it gets the point accross
function openOrLogLink(link:string) {
  if(autoOpenBrowser) {
    open(link)
    return;
  }
  logger.log(uiLink(link))
}
```

camden11: Hmm yeah that's a good callout. I think it _might_ be a little more complicated than that (and make make more sense as a follow up to this) because most of our `open` calls come after prompts, and we should avoid displaying the prompts if people have this set to off. But I like the idea of giving the config setting a more generic name so we can use it for other stuff in the future

camden11: Thinking about it more though, I kind of think that people might want the ability to turn this and only this off. It would be the only place where you're not given a prompt to open the browser, and we can handle those other cases individually. For example, if you want to install an OAuth app, you NEED to open the browser, so it's getting opened at some point whether it's automatic or not. We could update that prompt to log a link if you choose not to open it

joe-yeager (approved):

PR #295, opened 2025-08-07
Ts/fix module upload order batch meta

TanyaScales:
## Description and Context

Adjusts some of the code on the previous fix in isolation for testing -- Batch process on all meta.json files without checking for net new. Then process 'other' (calling deferred) files second in a different upload.


joeyblake (approved):
🚀

PR #307, opened 2025-09-22
Removing unnecessary behavior in the config file read util

brandenrodgers:
## Description and Context

From what I can tell there's no reason for us to be calling `isConfigPathInGitRepo` here. That util returns a boolean and always has, so it's doing nothing. [Here's the commit where it was initially adde for context](https://github.com/HubSpot/hubspot-cli/commit/b80acbfbe8ed2accf9d22771a27a13d8090ed78d#diff-d71973eeeaa7f10917f7688acc4c7e3d12d29743be4a92878758e199d89229d9R194). I'm guessing it was a mistake.


chiragchadha1 (approved):
🚢

PR #312, opened 2025-10-02
chore: Update release script

joe-yeager:
## Description and Context

Update to use the new npm-scripts package for releasing


lib/cms/uploadFolder.ts
```diff
@@ -191,7 +191,7 @@ async function uploadMetaJsonFiles(
   uploadFile: (file: string) => () => Promise<void>
 ): Promise<void> {
   const moduleMetaJsonFiles = moduleFiles.filter(isMetaJsonFile);
-  
+
```
chiragchadha1, line 194: is this intentional?

joe-yeager: Yeah, there was a prettier error otherwise

utils/cms/modules.ts
```diff
@@ -51,4 +51,3 @@ export function isModuleFolderChild(
     .slice(0, length - 1)
     .some(part => isModuleFolder({ ...pathInput, path: part }));
 }
-
```
joe-yeager, line 54: This was another prettier error

PR #393, opened 2026-04-01
feat: Implement project release API endpoints

chiragchadha1:
## Description and Context

Related to DPGDXFE-1041

Adds API functions and types for the new project release endpoints on `DFSBuildPipeline`'s `ProjectResource`. These are needed by the CLI's `hs project release` command family (DPGDXFE-1041, DPGDXFE-1042, DPGDXFE-1043).

Three new API functions:
- `createRelease(accountId, projectName, buildId)` — `POST /projects/{name}/releases`
- `listReleases(accountId, projectName, params?)` — `GET /projects/{name}/releases`
- `getReleaseInfo(accountId, projectName, releaseTag)` — `GET /projects/{name}/releases/{tag}`

Two new types in `types/Project.ts`:
- `Release` — release object with tag, buildId, createdAt, and optional components
- `FetchListReleasesResponse` — standard `CollectionResponse` with paging cursor

The release API is still being actively developed by the backend team and may change. This gets the foundation in place so the CLI can start building against it.

## TODO

- [ ] API shape may change as the backend iterates — types and functions will need updating

## Who to Notify


chiragchadha1:
Closing in favor of https://github.com/HubSpotEngineering/hubspot-cli-private/pull/669. Moving these endpoints directly into the CLI so we can iterate faster without cutting a local-dev-lib release for each API change. We'll move them back to local-dev-lib once the APIs stabilize.

PR #410, opened 2026-04-14
Support custom port manager port

camden11:
## Description and Context
The Port Manager previously always ran on port 8080, which is one of the most commonly used dev server ports. This resulted in users running into conflicts when running other local services simultaneously. This PR updates the Port Manager with a new default port, `4828`, and also adds a new option to run it on a custom port.

## Who to Notify
@brandenrodgers @joe-yeager @chiragchadha1


brandenrodgers (commented):
Will this require a major version bump?

brandenrodgers (approved):

camden11:
@brandenrodgers Good call, yeah I think removing the `PORT_MANAGER_SERVER_PORT` const would constitute a major bump. I think it might be worth keeping the old const even though it would be a little confusing. We can remove it whenever we have another major version planned.

chiragchadha1 (approved):

PR #427, opened 2026-05-01
feat: add log buffer + writeBufferedLogsToFile to logger

joe-yeager:
## Description and Context

Capture every call into an in-memory buffer (timestamped + level-tagged) regardless of the active LOG_LEVEL, so callers can flush the captured context after a failure.

Adds three new methods to the exported logger:
- viewBuffer(): peek at the joined buffer without clearing
- flushBuffer(): return the joined buffer and clear it
- writeBufferedLogsToFile({ dir, commandName, maxFiles? }): flush the buffer to a rotating log file under the consumer-provided dir. Caps the directory at maxFiles (default 3) by deleting oldest entries. Always clears the buffer, even on write failure.

### DIFF
--- a/lib/LogBuffer.ts
+++ b/lib/LogBuffer.ts
@@ -1,0 +1,117 @@
+import fs from 'fs';
+import path from 'path';
+
+const DEFAULT_BYTE_LIMIT = 128 * 1024 * 1024;
+const DEFAULT_MAX_LOG_FILES = 3;
+
+export type LogBufferOptions = {
+  byteLimit?: number;
+};
+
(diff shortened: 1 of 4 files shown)


lib/LogBuffer.ts
```diff
@@ -0,0 +1,116 @@
+import fs from 'fs';
```
joe-yeager, line 1: I split this into it's own class so we can use the same logic for other things such as tools logs, etc.

camden11 (approved):
Nice, LGTM. No need to add this now, but max number of log files could be a good option to add to config

PR #430, opened 2026-05-08
feat: add commerceLevel to DeveloperTestAccount type

juanfariasdev:
Added an optional `commerceLevel` property to the `DeveloperTestAccountConfig` type.


chiragchadha1 (approved):
Thanks for catching this and for creating the PR!

PR #443, opened 2026-06-05
feat: add installStaticAuthAppOnCurrentAccount

chiragchadha1:
## Description and Context

Adds `installStaticAuthAppOnCurrentAccount` to `api/appsDev.ts`, mirroring the existing `installStaticAuthAppOnTestAccount` but posting to the apps-hublets `/current-account` endpoint.

This backs the HubSpot CLI's `hs project install-app` flow for installing a project's static-auth app into the portal that **owns** it (the owner-portal path), alongside the existing test-account install. 

## Pre-review checklist

- [x] The `/ldl:code-check` skill has been run and the feedback has been addressed — lint clean; the one finding (no test coverage for `api/appsDev.ts`) is resolved below
- [x] Tests have been added for new behaviors — added `api/__tests__/appsDev.test.ts` with full coverage of all exported functions (8 tests)
- [x] Manually tested the changes — local-linked this branch's `dist` into the HubSpot CLI and verified the CLI builds, unit tests pass, and a live `hs project install-app` run resolves and uses this function end-to-end

## Screenshots

N/A — API client addition, no UI.

## Who to Notify
@joe-yeager @camden11

### DIFF
--- a/api/__tests__/appsDev.test.ts
+++ b/api/__tests__/appsDev.test.ts
@@ -1,0 +1,128 @@
+vi.mock('../../http');
+import { http } from '../../http/index.js';
+import { vi, describe, it, expect, afterEach } from 'vitest';
+import {
+  fetchPublicAppsForPortal,
+  fetchPublicAppDeveloperTestAccountInstallData,
+  fetchPublicAppProductionInstallCounts,
+  fetchPublicAppMetadata,
+  installStaticAuthAppOnTestAccount,
+  installStaticAuthAppOnCurrentAccount,
+  fetchAppMetadataByUid,
+  fetchAppMetadataBySourceId,
+} from '../appsDev.js';
+
+const APPS_DEV_API_PATH = 'apps-dev/external/public/v3';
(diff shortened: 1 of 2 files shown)


api/__tests__/appsDev.test.ts
```diff
@@ -0,0 +1,128 @@
+vi.mock('../../http');
```
chiragchadha1, line 1: I noticed there was no test file for `appsDev` so I decided to add it as part of this PR

api/appsDev.ts
```diff
@@ -60,6 +60,21 @@ export function installStaticAuthAppOnTestAccount(
   });
 }
 
+export function installStaticAuthAppOnCurrentAccount(
```
joe-yeager, line 63: Can we define a return type for the function?

api/__tests__/appsDev.test.ts
joe-yeager: Thanks for adding these

api/appsDev.ts
chiragchadha1: Added

camden11 (approved):

PR #446, opened 2026-06-12
fix: replace single-request folder tree fetch with incremental /meta/ traversal

joeyblake:
## Description and Context

`hs cms fetch` times out on directories with ~1300+ files. The root cause is that `downloadFolder` made a single `GET /filemapper/v1/download/{path}` call which required the server to serialize the entire recursive directory tree into one JSON response before returning. With enough files, this exceeds the HTTP timeout before the response completes.

Related issues on the public CLI repo: [#1575](https://github.com/HubSpot/hubspot-cli/issues/1575) and [#746](https://github.com/HubSpot/hubspot-cli/issues/746), both closed without a structural fix.

There was also a secondary bug: the `timeout: 60_000` passed by the `hs cms fetch` command was silently dropped by `getFileMapperQueryValues`, which only destructured `staging` and `assetVersion` from options — so the actual HTTP timeout was always the 15s default regardless of what the caller passed.

### Fix

Replace the single `/download/` call with an incremental walk using the existing `/meta/` endpoint (shallow listing) + `/stream/` (individual file download), which is already used by `hs cms list` and the VS Code extension.

**Before:**
downloadFolder → GET /download/{path} → entire tree in one JSON → times out

**After:**
downloadFolder → GET /meta/{path} → immediate children only
  → files: queue fetchAndWriteFileStream (PQueue, concurrency 10)
  → folders/modules: recurse into each subdirectory via /meta/

The `/meta/` endpoint returns a shallow listing (children as strings) and is fast — no server-side tree serialization.

### Research findings that shaped the implementation

- `/meta/` returns `children` as `string[]` (child names), not nested `FileMapperNode` objects. The TypeScript type was wrong; a separate `DirectoryMetaNode` type was added.
- `/meta/` works for `@hubspot` paths, but `/meta/` on `/` does NOT include `@hubspot` in children — it must be manually prepended, the same way the VS Code extension handles it.
- Module folders (`.module`) are treated as folders in the tree — `isPathToFile` returns false for them, so they recurse correctly.
- The module-level `PQueue` was shared across concurrent `downloadFolder` calls. One call's `onIdle()` would wait on another call's items; `clear()` would cancel unrelated work. The queue is now created inside `downloadFolder`, giving each download its own isolated instance.
- The old `success` variable was set inside async `queue.add` callbacks but read synchronously — the early break never actually fired. Replaced with a `Set<string>` of failed paths, read after `queue.onIdle()`.

### Design decisions

- **Subdirectory traversal is sequential** — `queueFolderTree` awaits each child folder's `/meta/` fetch and recursion before continuing. The single `PQueue` is used only for file downloads (concurrency 10); directory traversal is not parallelized.
- **`queue.onIdle()` called only once** — at the outermost `downloadFolder` level, after the full tree walk completes. `queueFolderTree` never waits on the queue.

### Open questions / known risks

- **Do `/meta/` query params work the same as `/download/`?** The `buffer` and `environmentId` params are now passed to `/meta/` calls. If the endpoint doesn't support them, draft mode fetching may behave differently. Basic testing looked fine but edge cases around staging/draft mode should be verified.
- **Performance for deep trees** — Sequential `/meta/` traversal for deeply nested directories adds latency before file downloads begin. Acceptable for now; subdirectory traversal could be parallelized in a follow-up.

## Pre-review checklist

## Screenshots

N/A — no UI changes.

### DIFF
--- a/api/fileMapper.ts
+++ b/api/fileMapper.ts
@@ -148,7 +148,7 @@ export function getDirectoryContentsByPath(
   });
 }
 
-// Get shallow directory listing — children are strings, not nested nodes
+// getDirectoryContentsByPath has the wrong return type for /meta/ — children are strings, not FileMapperNodes
 export function getDirectoryMetaByPath(
   accountId: number,
   path: string,
--- a/api/fileMapper.ts
+++ b/api/fileMapper.ts
@@ -155,7 +155,7 @@ export function getDirectoryMetaByPath(
   options: FileMapperOptions = {}
 ): HubSpotPromise<DirectoryMetaNode> {
   return http.get<DirectoryMetaNode>(accountId, {
-    url: `${FILE_MAPPER_API_PATH}/meta/${encodeURIComponent(path)}`,
+    url: `${FILE_MAPPER_API_PATH}/meta/${path.split('/').map(encodeURIComponent).join('/')}`,
     ...options,
   });
 }
--- a/api/fileMapper.ts
+++ b/api/fileMapper.ts
@@ -155,7 +155,7 @@ export function getDirectoryMetaByPath(
   options: FileMapperOptions = {}
 ): HubSpotPromise<DirectoryMetaNode> {
   return http.get<DirectoryMetaNode>(accountId, {
-    url: `${FILE_MAPPER_API_PATH}/meta/${path}`,
+    url: `${FILE_MAPPER_API_PATH}/meta/${encodeURIComponent(path)}`,
     ...options,
   });
 }
--- a/api/fileMapper.ts
+++ b/api/fileMapper.ts
@@ -5,7 +5,12 @@ import { AxiosResponse } from 'axios';
 import { http } from '../http/index.js';
 import { getCwd } from '../lib/path.js';
 import { FILE_MAPPER_API_PATH } from '../constants/endpoints.js';
-import { FileMapperNode, FileMapperOptions, FileTree } from '../types/Files.js';
+import {
+  DirectoryMetaNode,
+  FileMapperNode,
+  FileMapperOptions,
+  FileTree,
+} from '../types/Files.js';
 import { HubSpotPromise } from '../types/Http.js';
 
 export function createFileMapperNodeFromStreamResponse(
@@ -140,5 +145,17 @@ export function getDirectoryContentsByPath(
 ): HubSpotPromise<FileMapperNode> {
   return http.get(accountId, {
     url: `${FILE_MAPPER_API_PATH}/meta/${path}`,
+  });
+}
+
+// Get shallow directory listing — children are strings, not nested nodes
+export function getDirectoryMetaByPath(
+  accountId: number,
+  path: string,
+  options: FileMapperOptions = {}
+): HubSpotPromise<DirectoryMetaNode> {
+  return http.get<DirectoryMetaNode>(accountId, {
+    url: `${FILE_MAPPER_API_PATH}/meta/${path}`,
+    ...options,
   });
 }
--- a/lib/__tests__/fileMapper.test.ts
+++ b/lib/__tests__/fileMapper.test.ts
@@ -422,7 +422,9 @@ describe('lib/fileMapper', () => {
 
       await downloadFileOrFolder(testAccountId, 'ThemeDirectory/', './');
 
-      const metaCalls = getDirectoryMetaByPath.mock.calls.map(c => c[1] as string);
+      const metaCalls = getDirectoryMetaByPath.mock.calls.map(
+        c => c[1] as string
+      );
       expect(metaCalls.some(p => p.includes('//'))).toBe(false);
       expect(metaCalls).toContain('ThemeDirectory/modules');
     });
--- a/lib/__tests__/fileMapper.test.ts
+++ b/lib/__tests__/fileMapper.test.ts
@@ -412,14 +412,6 @@ describe('lib/fileMapper', () => {
         )
         .mockResolvedValueOnce(
           mockAxiosResponse({
-            name: 'ThemeDirectory',
-            path: '/ThemeDirectory',
-            folder: true,
-            children: ['modules'],
-          })
-        )
-        .mockResolvedValueOnce(
-          mockAxiosResponse({
             name: 'modules',
             path: '/ThemeDirectory/modules',
             folder: true,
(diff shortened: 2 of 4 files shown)


lib/__tests__/fileMapper.test.ts
```diff
@@ -290,4 +291,142 @@ describe('lib/fileMapper', () => {
...
+          folder: true,
+          children: ['bad.html'],
+        })
+      );
+      pathExistsSpy.mockImplementation(async () => false);
+      fetchFileStream.mockRejectedValue(new Error('network error'));
+
+      await expect(
+        downloadFileOrFolder(testAccountId, '/templates', './')
+      ).rejects.toThrow();
```
chiragchadha1: I think it'd make sense to test specifically for the incompleteFetch error. And if possible, check to see that multiple queued files continue to download even if one fails (i.e. `bad.html` fails but `good.html` still runs)

lib/__tests__/fileMapper.test.ts
```diff
@@ -290,4 +291,142 @@ describe('lib/fileMapper', () => {
...
+    });
+
+    it('wraps /meta/ failures as failedToFetchFolder', async () => {
+      getDirectoryMetaByPath.mockRejectedValue(new Error('server error'));
+
+      await expect(
+        downloadFileOrFolder(testAccountId, '/templates', './')
+      ).rejects.toThrow();
```
chiragchadha1: Check specifically for `failedToFetchFolder` here

api/fileMapper.ts
```diff
@@ -142,3 +147,15 @@ export function getDirectoryContentsByPath(
...
 }
+
+// getDirectoryContentsByPath has the wrong return type for /meta/ — children are strings, not FileMapperNodes
+export function getDirectoryMetaByPath(
+  accountId: number,
+  path: string,
+  options: FileMapperOptions = {}
+): HubSpotPromise<DirectoryMetaNode> {
+  return http.get<DirectoryMetaNode>(accountId, {
+    url: `${FILE_MAPPER_API_PATH}/meta/${encodeURIComponent(path)}`,
```
chiragchadha1: Encoding the full path would result in `/meta/ThemeDirectory%2Fmodules` instead of `/meta/ThemeDirectory/modules` — this is different than what `getDirectoryContentsByPath` does. Can you confirm this is correct?

lib/fileMapper.ts
```diff
@@ -343,69 +298,123 @@ export async function fetchFolderFromApi(
...
-          version: getAssetVersionIdentifier(options.assetVersion, src),
-          dest,
-        })
-      );
-    } else {
-      // TODO: Fix this exception. It is triggering the catch block so this error is being rewritten
-      throw new Error(i18n(`${i18nKey}.errors.incompleteFetch`, { src }));
-    }
   } catch (err) {
+    queue.clear();
```
chiragchadha1, line 426: This would only clear pending downloads, not ones that have already started. Do we want to exit immediately here?

lib/fileMapper.ts
```diff
@@ -343,69 +298,123 @@ export async function fetchFolderFromApi(
...
+  if (isRoot) {
+    children = ['@hubspot', ...children];
+  }
+
+  await fs.ensureDir(localPath);
+  logger.log(i18n(`${i18nKey}.wroteFolder`, { filepath: localPath }));
+
+  const queryValues = getFileMapperQueryValues(cmsPublishMode, options);
+
+  for (const childName of children) {
```
chiragchadha1, line 354: If a child is a dotted folder name like `v1.0`, it would be incorrectly classified as a file with `isPathToFile(childRemotePath)`. Are the child names guaranteed to not include dotted folder names?

lib/fileMapper.ts
```diff
@@ -343,69 +298,123 @@ export async function fetchFolderFromApi(
...
+): Promise<void> {
+  if (!directoryNode.folder) return;
+
+  const { isRoot } = getTypeDataFromPath(src);
+  let children: string[] = directoryNode.children || [];
+  if (isRoot) {
+    children = ['@hubspot', ...children];
+  }
+
+  await fs.ensureDir(localPath);
```
chiragchadha1: If this fails it'll be caught by `downloadFolder()` and surfaced as a Design Manager fetch failure. Could we preserve the filesystem error wrapping here so local write failures point users at the right cause?

lib/fileMapper.ts
```diff
@@ -343,69 +298,123 @@ export async function fetchFolderFromApi(
   return node;
 }
 
+async function queueFolderTree(
```
chiragchadha1, line 304: Can we use an options arg here?

chiragchadha1 (commented):
Overall this looks great (still need to test though)! A few comments. Also, can we increase testing coverage (specifically on the recursive file downloading and `/meta/` path creation to ensure it's working properly?

lib/fileMapper.ts
joeyblake: updated

lib/fileMapper.ts
joeyblake: Great catch! I double checked this with the api, and that would have been a problem this is fixed now.

lib/fileMapper.ts
joeyblake: We could, but if we do that wouldn't aborting those mid-write would leave partial files on disk? Which probably is worse than letting them finish. Or we could add the complexity of threading an AbortController through and cleaning things up, I'm not sure if the juice is worth the squeeze there. If there are failures and the command just runs again, it by default skips the files that have been completed. In my time using the cli to move files around I haven't seen other cases be that thorough. What do you think?

chiragchadha1: That makes sense! Let's leave it as is then

> I'm not sure if the juice is worth the squeeze there.

ha

lib/fileMapper.ts
```diff
@@ -56,6 +54,10 @@ export function isPathToRoot(filepath: string): boolean {
   return /^(\/|\\)?$/.test(filepath.trim());
 }
 
+export function isPathToFolder(filepath: string): boolean {
```
chiragchadha1: This is being used anywhere, do we need it?

joeyblake: no, dropped it.

joe-yeager (approved):

chiragchadha1 (approved):

PR #467, opened 2026-07-24
feat: add getBaseUrl util to simplify website origin lookups

ashleighliu:
## Description and Context

- Adds a `getBaseUrl(accountId)` utility in `lib/urls.ts` that wraps the verbose `getHubSpotWebsiteOrigin(getConfigAccountEnvironment(...))` pattern into a single call
- Updates `lib/cms/themes.ts` to use the new util, replacing the inline baseUrl construction
- This util can also be used in `hubspot-cli-private` to simplify the same repeated pattern there

Towards https://linear.app/hubspot/issue/DPGDXFE-63/replace-baseurl-creation-with-a-call-to-a-library-wrapper-function

## Pre-review checklist

### DIFF
--- a/lib/urls.ts
+++ b/lib/urls.ts
@@ -12,7 +12,7 @@ function getEnvUrlString(env?: string): string {
 export const getHubSpotWebsiteOrigin = (env: string) =>
   `https://app.hubspot${getEnvUrlString(env)}.com`;
 
-export function getBaseUrl(accountId: number): string {
+export function getHubSpotWebsiteOriginByAccountId(accountId: number): string {
   return getHubSpotWebsiteOrigin(
     getConfigAccountEnvironment(accountId) === ENVIRONMENTS.QA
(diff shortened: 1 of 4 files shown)


lib/urls.ts
```diff
@@ -11,6 +12,14 @@ function getEnvUrlString(env?: string): string {
 export const getHubSpotWebsiteOrigin = (env: string) =>
   `https://app.hubspot${getEnvUrlString(env)}.com`;
 
+export function getBaseUrl(accountId: number): string {
```
brandenrodgers: Slight nit, but can you clarify in the naming of this util that it's specifically for getting the website url and not the api one? Just to cut down on confusion.

ashleighliu: Makes sense, just changed to `getHubSpotWebsiteOriginByAccountId`, let me know if I can clarify further!

brandenrodgers (approved):
lgtm!

PR #472, opened 2026-08-11
feat: add isGithubError error guard

ashleighliu:
## Description and Context

The HubSpot CLI is adding failure tracking to `hs project create` and needs to tell GitHub-origin errors apart from other failures. The existing guards (`isSpecifiedError`, `isGithubRateLimitError`) only recognize `HubSpotHttpError`, but the GitHub API layer (`api/github.ts`) uses plain `axios`, so real GitHub failures arrive as raw `AxiosError`s that those guards miss. 

Adds `isGithubError(err)` in `errors/index.ts`, detecting GitHub origin on either error shape:
- `x-github-request-id` response header — present on every GitHub response, so it works for any GitHub HTTP failure (404, 403, 5xx), on both `HubSpotHttpError` (`.headers`) and `AxiosError` (`.response.headers`).
- Request URL host (`error.config.url` contains `github.com` / `githubusercontent.com`) as a fallback for network errors that have no response.

Towards https://linear.app/hubspot/issue/DPGDXFE-62/add-granular-tracking-for-hs-project-create-errors

## Pre-review checklist

## Who to Notify

@HubSpotEngineering/dev-experience-fe

### DIFF
--- a/errors/index.ts
+++ b/errors/index.ts
@@ -1,4 +1,4 @@
-import { AxiosError } from 'axios';
+import { isAxiosError } from 'axios';
 import {
   HubSpotHttpError,
   HubSpotHttpErrorName,
@@ -97,7 +97,7 @@ export function isGithubError(err: unknown): boolean {
   if (isHubSpotHttpError(err)) {
     return !!err.headers && 'x-github-request-id' in err.headers;
   }
-  if (err instanceof AxiosError) {
+  if (isAxiosError(err)) {
     const headers = err.response?.headers;
     if (headers && 'x-github-request-id' in headers) {
       return true;
--- a/errors/index.ts
+++ b/errors/index.ts
@@ -1,3 +1,4 @@
+import { AxiosError } from 'axios';
 import {
   HubSpotHttpError,
   HubSpotHttpErrorName,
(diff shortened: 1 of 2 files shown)


errors/index.ts
```diff
@@ -92,6 +93,24 @@ export function isGithubRateLimitError(err: unknown): err is HubSpotHttpError {
   );
 }
 
+export function isGithubError(err: unknown): boolean {
+  if (isHubSpotHttpError(err)) {
+    return !!err.headers && 'x-github-request-id' in err.headers;
+  }
```
camden11, line 99: Is there a scenario where an error can be a `HubSpotHTTPError` AND a GithubError?

errors/index.ts
```diff
@@ -92,6 +93,24 @@ export function isGithubRateLimitError(err: unknown): err is HubSpotHttpError {
   );
 }
 
+export function isGithubError(err: unknown): boolean {
```
camden11, line 96: Is there a standard shape we expect for errors from github? If so, we could use a type guard here (`arg is GithubError`) similar to how we do it for `HubSpotHTTPError`

ashleighliu: Did some research, unfortunately the answer is no. In general, the GitHub errors will come in as `AxiosError`, but could occasionally be `HubSpotHTTPError` as well. After doing some homework, though, there is a type for the body of GitHub errors that I could add:
```ts
export type GithubApiErrorBody = {
  message: string;
  documentation_url?: string;
  status?: string;
  errors?: Array<{
    resource?: string;
    field?: string;
    code?: string;
    message?: string;
  }>;
};
```
Though I can't really use this here since this type is the body of the error and not the error itself

errors/index.ts
ashleighliu: Yes, there exists a scenario in which this could happen: a GitHub fetch routed through a HubSpot service. This call goes through the wrapped HTTP client, which produces a `HubSpotHttpError`. If the proxy forwards GitHub's `x-github-request-id` response header, the resulting error is both a `HubSpotHttpError` and a GitHub error, hence I have this additional logic above

camden11 (dismissed):
That all makes sense, thanks for answering my questions! I think you can probably remove the `GithubApiErrorBody` type since it's not used (unless you have a planned use for it in the CLI). Otherwise LGTM

ashleighliu:
> That all makes sense, thanks for answering my questions! I think you can probably remove the `GithubApiErrorBody` type since it's not used (unless you have a planned use for it in the CLI). Otherwise LGTM

Np!! I added it for future use in cases where that body is needed, but no plans to use it yet so I'll just leave it out for now 👍

camden11 (approved):

PR #476, opened 2026-08-20
refactor: make GitHub error guards cause-aware and axios-aware

ashleighliu:
## Description and Context
Consolidates GitHub error-classification logic into local-dev-lib. Follow-up to #472.

- `isGithubRateLimitError` now also handles raw `AxiosError` (matching its sibling `isGithubError`) and looks through `error.cause`.
- `isGithubError` now looks through `error.cause`.
- Adds `getHttpStatusFromError` to pull an HTTP status off `HubSpotHttpError` or `AxiosError`, unwrapping `cause`.

The cause-unwrapping lives here because this library is what re-throws GitHub failures wrapped in `new Error(msg, { cause })` (see `lib/github.ts`), so its own guards should see through that.

Towards https://linear.app/hubspot/issue/DPGDXFE-62/add-granular-tracking-for-hs-project-create-errors

## Pre-review checklist

## Who to Notify

@HubSpotEngineering/dev-experience-fe

### DIFF
--- a/errors/index.ts
+++ b/errors/index.ts
@@ -82,7 +82,18 @@ export function isHubSpotHttpError(error?: unknown): error is HubSpotHttpError {
   );
 }
 
-export function isGithubRateLimitError(err: unknown): boolean {
+export function isGithubRateLimitError(err: unknown): err is HubSpotHttpError {
+  if (!isHubSpotHttpError(err)) {
+    return false;
+  }
+  return (
+    !!err.headers &&
+    err.headers['x-ratelimit-remaining'] === '0' &&
+    'x-github-request-id' in err.headers
+  );
+}
+
+export function hasGithubRateLimitError(err: unknown): boolean {
   return (
     hasGithubRateLimitSignal(err) ||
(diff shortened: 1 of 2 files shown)


errors/index.ts
```diff
@@ -82,18 +82,40 @@ export function isHubSpotHttpError(error?: unknown): error is HubSpotHttpError {
   );
 }
 
-export function isGithubRateLimitError(err: unknown): err is HubSpotHttpError {
```
camden11, line 85: Changing the return type of `isGithubRateLimitError` is technically a breaking change, which we can't do without a major version bump and a bit of advance planning.

I think adding an overload signature like this should resolve that

```
export function isGithubRateLimitError(err: HubSpotHttpError): err is HubSpotHttpError;
export function isGithubRateLimitError(err: unknown): boolean;
```

ashleighliu: Good catch! I think unknown callers still resolve to the boolean overload, so I left this as-is and added a new function `hasGithubRateLimitError` instead, let me know if that still works! 🙏

joe-yeager (approved):
LGTM

